PR #6, opened 2018-05-20
Remove the word "simply" from documentation

adactio:
I know the intention is to make the process seem straightforward, but using words like "simply" or "just" can have the unintended effect of making the reader feel inadequate.

http://bradfrost.com/blog/post/just/

https://adactio.com/journal/13521

Hope that helps,

Jeremy


gmrchk:
Thank you for your input.

PR #16, opened 2018-06-14
Add new skipPopStateHandling option for dynamic skip of popState even…

JonathanTron:
I needed a way to dynamically choose is swup should handle the popState itself or just ignore it.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -45,6 +45,8 @@ export default class Swup {
             disableIE: false,
             plugins: [],
 
+            skipPopStateHandling: function(){ return false; },
+
             LINK_SELECTOR: 'a[href^="/"]:not([data-no-swup]), a[href^="#"]:not([data-no-swup]), a[xlink\\:href]'
         }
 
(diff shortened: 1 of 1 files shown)


gmrchk:
Can you please elaborate on the use case for this?

JonathanTron:
Hi @gmrchk, 

I have content on some pages which need to handle the popState themselves, for instance impress.js. Using this hook, I can then peek and choose (in my case depending on what the url looks like) if it should trigger a swup transition or not.

Hope this clarify my need for this and thank you for the amazing library!

gmrchk:
Thanks!

I've never used swup with some other library using history API and I see how this could be a problem. Good job! 

I have merged your proposal and then modified the behavior a bit:
1. popstate event is passed into the function so users have all the available information
2. swup now adds property `source` to the history state objects it creates 
3. this property is used by default in the `skipPopStateHandling` function for the decision ([check here](https://github.com/gmrchk/swup#skip-popstate-handling))

Please, let me know if you see any issue with this approach. I could be missing something, as I've never used swup and some other history API library together.

JonathanTron:
Thanks for the merge.

Your changes looks good to me, far more complete than what I did.

PR #28, opened 2018-08-12
Update README example with correct DOM structure

dylanfisher:
It seems like the DOM elements are not nested properly in the README example.


gmrchk:
It doesn't really matter to swup what is the structure, both works... you could actually event place both, id and animation class, to the same element. Two elements are used in the readme only to demonstrate that it can be different elements. However, I get your point. It looks kind of weird.

Cheers!

Issue #47, opened 2018-09-22
Option to not skip animations when going back in history


OSDVF:
The default swup behaviour is skipping animations when you go back in history.
I tried to animate the pages by assigning to options.skipPopStateHandling a function which always returns true (prevents the default behaviour) and loads the page by loadPage() so it does animate.
But this also creates a new entry in browser history and when you now try to go back it loads the same page over and over.
So the only way would be altering the Swup files.
Thank you for taking this into consideration

jcklpe:
@OSDVF 

I just got swup working with the wordpress theme I'm working on and I realized that it was having this behavior at almost the exact same time as you. The animation is now working beautifully but when I hit back, no more smooth transition! Would love a solution! Thanks!

jcklpe:
I don't think if there is a single "right" choice for back button behavior. I think a lot of it is going to be dependent upon the particular design of the website, expectations of the user which will be related to target audience, etc. So I think your points are certainly fine. I also don't have an iPhone so it's hard for me to comment as I don't have experience with that design pattern. 

I think your point on hitting back and losing the scroll position is valid. That is pretty odd. 

I don't know if there is anything inherently bad about not going back immediately and having the animation. I think there are trade offs there. On one hand it might be slightly slower, on the other hand it might be seen as a more consistent thread of how action and feedback are connected in the user's mental model of the interaction. 

Personally for me, one reason why I would want the consistency between the two is because the website I'm building visual elements that I want to see continuously represented between pages. So for instance on the index.php I have a section to click to see a post page. The background of the section is an image that is also the Hero Image at the top of the page. Having a consistent transition back and forth between the two helps to cinch the illusion of continuous space. 

You can kind of see this treatment of the hero image as continuous looking at this page here: https://tympanus.net/Development/GridLayoutScrollableContent/ Though there it's being used in a modal context so the back button is not enabled. But I think you can see what I mean if you go back and forth between the expanded and unexpanded states of the posts.

OSDVF:
To be more specific, I want to animate every transition between pages because I´m writing a PWA which is intended to run as an "native" app on Android and iOS. All apps have animations when they switch pages/activities/etc.. so I think it would be very ugly to not have them ;)

panayotov:
I ran into the same issue and solved it in a little bit odd way,
I moved the transition from animations object to each page. My code is a little bit architecture-specific, but basically this is what I'm doing:

1) Every page has 'data-page' attribute that acts like Barbajs namespace;
[...]

PR #56, opened 2018-09-28
Add animateHistoryBrowsing support

OSDVF:
An attempt to add an "animateHistoryBrowsing" option to make animations work not only when user navigates through links but also using back/forward buttons. May be useful for PWAs etc.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -49,6 +49,7 @@ export default class Swup {
                 }
                 return true;
             },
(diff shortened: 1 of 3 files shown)


gmrchk:
implements #47

gmrchk:
Good job on this. I've made a few minor changes and updated readme accordingly.

PR #69, opened 2018-10-18
Document "X-Requested-With" header

alwaysblank:
This documents the header swup sets on requests, which allows user to customize the content returned. This seems like useful information for optimizing requests, and I figured it might save someone like me 5 minutes of digging through the source code in the future.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -44,6 +44,8 @@
 
 [API](#api)
 
+[Request Header](#request-header)
+
 [Contributions](#contributions)
 
 ## Installation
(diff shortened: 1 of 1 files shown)


gmrchk:
Hi @alwaysblank, good job on this! Totally forgot someone might want to know this.

I've modified your contribution to include some (I think) valuable information. Hope you don't mind.

Cheers!

alwaysblank:
I don't mind at all, thank you! That wasn't something that had occurred to me (for my use case it didn't matter: I was only modifying stuff outside of the part of the page swup looks at). 

Thanks for all your hard work on this library, it's amazing and I'm really enjoying working with it. :)

PR #74, opened 2018-10-30
Fix FireFox right click bug

drivebass:
On Firefox a right click over a link triggers 'clickLink' event. Check if it's a right click to prevent the trigger. Doesn't work with event.which check, maybe with contextmenu event somewhere?

### DIFF
--- a/dist/swup.js
+++ b/dist/swup.js
@@ -526,7 +526,7 @@ var Swup = function () {
         key: 'linkClickHandler',
         value: function linkClickHandler(event) {
             // no control key pressed
-            if (!event.metaKey) {
+            if (!event.metaKey && !ev.which == 3) {
                 this.triggerEvent('clickLink', event);
                 var link = new _Link2.default();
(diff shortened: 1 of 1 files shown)


gmrchk:
Hi @drivebass, thank you for your input! 

I really appreciate your effort, but if you'd like to contribute, you cannot implement your changes in the compiled version of swup (dist folder).

As this is a one-liner, I have fixed this elsewhere using `event.button`, which should work as expected.

drivebass:
Hi @gmrchk, I have to say this is the most complete and smooth plugin for history transitions by far! Sorry if I messed up, not very used with github's way of pull-requests and etc and I didn't realize the changes were implemented in the compiled version. Big thanks for letting me know, the fix works great on FF now!

gmrchk:
Hi @drivebass, thanks for the nice words! And no worries, I'm glad people are trying to help, even if it doesn't work out every time.

PR #75, opened 2018-11-02
Fix missing serverError handler list

wongjn:
If Swup catches a `500` response, it does not fallback to conventional navigation since it errors in a way that is uncatchable from the promise. This is because it errors in `triggerEvent` because `this._handlers.serverError` does not exist.


gmrchk:
Don't know how I missed that... thanks!

PR #77, opened 2018-11-05
Add some default values to readme

rbuchberger:
This is a super minor change, but it tripped me up for a minute. It adds default values to the descriptions of 2 properties of the options object, so they're consistent with the others.


gmrchk:
Sounds good, thanks!

PR #94, opened 2019-01-05
Add Accept text/html header to request

sw-double:
This prevents issues with browsersync proxy, it does not rewrite links hosts in html without this header

Turbolinks [does this](https://github.com/turbolinks/turbolinks/blob/c340c42e2ece8c057aec611074971462a6e76c71/src/turbolinks/http_request.coffee#L66) as well

### DIFF
--- a/src/modules/request.js
+++ b/src/modules/request.js
@@ -24,6 +24,7 @@ module.exports = function(options, callback = false) {
 
     request.open(data.method, data.url, true)
     request.setRequestHeader("X-Requested-With", "swup")
+    request.setRequestHeader("Accept", "text/html, application/xhtml+xml")
     request.send(data.data)
     return request
 }


sw-double:
I think this library only makes requests for html pages but not images or any other stuff

gmrchk:
I've been using swup with browsersync quite frequently and never had such issue. Could you elaborate on when is this a problem? 

Could the issue be more related to the response from the server, eg. not setting `Content-Type` correctly or something like this?

When extending the swup and modifying it to your needs (primarily by rewriting the getDataFromHtml method), swup can also accept other types like JSON. In that case, rewriting of this line would be also necessary, which I would like to avoid if it's not necessary.

MartDSam:
Here I am also having problem with browserSync. I am using the proxy option and when the swup plugin of the merge in the head it pulls the original link from the site and not the proxy link.

The original project link is client.localhost: 8000, the proxy link is client.localhost: 3000. But when the swup of the merge in the head the link of the css is with the link of the port of 8000 and not of 3000.

gmrchk:
Request headers can be modified with options in v2. 

Thank you, everyone, for the cooperation.

PR #151, opened 2019-06-03
feat: support body container

Grafikart:
Accept body as a container


Grafikart:
I also added the page to the loadedEvent since it can be used to add more behaviours

gmrchk:
Hi, swup relies on body element in some cases, like the [Body Class plugin](https://swup.js.org/plugins/body-class-plugin) or for themes as well, where replacing the whole body would break it all (at least for themes for sure), so those would be incompatible when this option would be set to `body`. Instead of replacing the whole body, swup goes for replacing any parts of the page, even if it means the whole page being wrapped in one container. 

I think this would complicate thing in some areas, and quite unnecessarily. Do you have some specific example where the whole body just needs to be replaced, and classic containers can't be used?

gmrchk:
Kudos for the `pageLoaded` event tho, could be very useful in some cases.

gmrchk:
Unfortunately, I'll have to close this as it's too much of a breaking change without actual feedback of this being requested by people. I have updated the docs to include a warning.

Thank for all your work!

Issue #145, opened 2019-05-31
Build script fails due to outdated webpack-cli version


tedw:
Need to update to webpack-cli 3.x to fix the build script (see https://github.com/webpack/webpack-cli/issues/607)

PR #157, opened 2019-06-11
Upgrade webpack-cli to fix build script error #145

tedw:


gmrchk:
Totally missed I got the different version installed globally. Good catch!

Issue #146, opened 2019-05-31
Error in IE 10 due to lack of dataset support


tedw:
Need to swap out [`dataset`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset) for `getAttribute()`/`setAttribute()` to fix errors in IE 10 ([caniuse.com](https://caniuse.com/#feat=dataset&search=dataset)).

PR #158, opened 2019-06-11
Fix IE10 errors by replacing dataset with get/setAttribute() #146

tedw:


gmrchk:
Hi, sorry for the delay.

I would usually go for doing this in a different way like add a polyfill or transpile it during the build. 

However, since I find the `getAttribute`, `setAttribute`. `removeAttribute` way more readable than a previous `dataset` stuff, let's get rid of two problems at once and go with your solution. 

Good job and thanks for your time! 👏

PR #214, opened 2019-10-16
Feature/dynamically add aria live attribute

MMasey:
Hi @gmrchk, this PR add's the `aria-live="polite"` attribute to the body element when swup is initialised. After playing around with the  `aria-live="assertive"` and `aria-atomic="true"`, I've found the proposed solution to be the most ideal.

I've tested it a little bit, but I imagine you'll want to run your own tests on it too to make sure you're happy. It's a pretty small change, but it should help to improve the accessibility.

Many thanks

P.S. I also did an `npm audit fix` to reduce the number of warnings when doing the `npm install`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "swup",
-  "version": "2.0.6",
+  "version": "2.0.7",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


gmrchk:
Hi @MMasey, sorry it took a while. Good work on the PR, thank you for your help!

I've looked through some sources like [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) and [other](https://bitsofco.de/using-aria-live/) and it seems that adding `aria-live` after the initial load won't make any difference. 

> Dynamically adding an element with an aria-live attribute or specialized role to the document also won't result in any announcement by assistive technologies (as, at that point, the browser/assistive technologies are not aware of the live region yet, so cannot monitor it for changes).

I think the only option to do this correctly is by adding the attributes to the swup containers manually. I would propose to add the info about using `aria-live` attributes to the documentation. What do you think?

Good point with the `npm audit fix`. 👍 Will do that in a separate commit.

PR #260, opened 2020-04-28
Fix bug with back button error redirects

jaredcwhite:
Swup is so great, I've really enjoyed being able to use it on https://www.bridgetownrb.com!

However, I ran into an issue where I have two different page templates with a different number of containers. I was able to get things working all right with a bit of clever contextual setup in DOMContentLoaded—a full page reload is triggered going from one template to another, but within that template all the different pages transition well, so that's all fine.

BUT, when trying to hit the back button from a page using Template B back to a page using Template A, the browser was not going back and kept redirecting forward to the Template B page again. Super annoying. After some digging around, I pinpointed the part of the code that was causing the problem. Essentially after the first `window.history.go(-1)` to reload the new Template B page, after browsing around and then using the back button to go back to a Template A page, this popstate handler would do the same reload of Template B page, which *isn't* what is needed in that context.

Anyway, this is a bit hard to describe, but I did verify this fix solved the problem on the site mentioned above, so hopefully you'll find it useful. Let me know if you have any questions.


jaredcwhite:
@gmrchk Any chance for a review of this? Thanks :)

hirasso:
Sorry @jaredcwhite for this very late reaction. I'm in the process of going through open PRs and Issues right now and stumbled upon this old gem 💎

I just had a look at your site https://www.bridgetownrb.com/. It seems like you are now using [hotwire/turbo](https://turbo.hotwired.dev/), so probably you don't have a little demo of the described issue at hand, anymore? 

Anyways, I think I am getting what you mean. I will set up a test and apply your code.

hirasso:
Ah, just discovered [your related PR](https://github.com/bridgetownrb/bridgetown/pull/471) for moving to turbo. I will close this PR then but will open another one, as soon as I find the time to test your code.

daun:
@hirasso I've tested this locally and I couldn't reproduce the error in the first place. It might have something to do with a recent fix to the preload plugin, but I couldn't say if the PR solves an actual issue.

hirasso:
I just set up a test site and noticed that there is definitely something odd going on if there is a containers mismatch between pages or if there is no container in the response, at all. I will investigate further and try to create a PR with fixes.

How I see it, we could handle this gracefully without adding a lot of complexity by rejecting the `xhrPromise` and triggering a hard page reload, if a response's `page.blocks` has a length of `0`.

Two other things I noticed while digging through the code: 

- I would opt for always use `console.warn` instead of `console.error`, swup is meant to be a progressive enhancement, after all. 
- I would like log messages to be prefixed with `[swup]` or the like, so `Container #swup not found on page.` would become `[swup] Container #swup not found on page.`, for example.

Do you agree, @daun ?

hirasso:
Sorry, got a bit confused here. I also can't reproduce the exact issue descibed above anymore. Found another one, though. Closing this again.

PR #356, opened 2021-04-10
Replace unsupported dependency

fregante:
`delegate` has several bugs and it hasn't been updated since 2017: https://github.com/zenorocha/delegate

I forked it a few years ago and modernized it along the way: https://github.com/fregante/delegate-it

Being a fork, the API is identical, but browser compatibility now lacks IE due to a `.closest` and `WeakMap`


daun:
@gmrchk How do you feel about this reducing browser support to exclude IE11? Is there a support matrix somewhere that you're aiming for? It'd be great not to rely on outdated dependencies, not least for security reasons.

fregante:
Ping on this if you're interested

daun:
We finally compiled a current [browser support matrix](https://swup.js.org/other/browser-support). Looks like switching to `delegate-it` won't change the supported browsers. IE11 isn't supported without a Promise polyfill, and `WeakMap` is supported in IE11 and Edge 12.

I'll look into merging the PR soon after doing some testing.

daun:
Added a few more tests, looking good to merge soon. Thanks for the PR!

daun:
Merged into master as 2.0.15. Thanks again!

PR #375, opened 2021-07-21
Fix invalid selector in samePageWithHash

knokmki612:
currently not considered the following cases in samePageWithHash event.

- include url encoded characters
- include ':'

this cause the DOMException as follows when click an anchor link something like `#テスト` (this sample is using swup with scroll-plugin):

```
DOMException: Document.querySelector: '#%E3%83%86%E3%82%B9%E3%83%88' is not a valid selector swup.min.js:1:16887
Uncaught DOMException: Document.querySelector: '#%E3%83%86%E3%82%B9%E3%83%88' is not a valid selector swup.min.js:1
```

this change will fix the above exception and improve to keep valid dom element extraction.


daun:
Thanks for your contribution!

PR #440, opened 2022-06-27
Fix anchor selector

daun:
Part 1 of an effort to improve swup finding anchors to scroll to.  
Part 2 will be a PR on the scroll plugin.

- Create tests for anchors swup currently fails to find:
  - `<a name="anchor-by-name"></a>`
  - `<a id="anchor:with:colon"></a>`
  - `<a id="anchor-with-unicode-テスト"></a>`
- Fix anchor selector by escaping CSS identifiers
- Speed up tests by removing timeouts where possible


daun:
@gmrchk  The new tests are failing in CircleCI since they require the second PR on the scroll plugin to be merged and released. Can you merge into master anyway? The current swup behavior isn't changed by this PR (hence the new failing tests).

src/utils/index.js
```diff
@@ -13,3 +13,13 @@ export const queryAll = (selector, context = document) => {
...
+
+const escapeCssRegex = /(^[^_a-zA-Z\u00a0-\uffff]|[^-_a-zA-Z0-9\u00a0-\uffff])/g;
+
+export const escapeCssIdentifier = (ident) => {
+	if (window.CSS && window.CSS.escape) {
+		return CSS.escape(ident);
+	} else {
+		return ident.replace(escapeCssRegex, "\\$1");
+	}
+};
```
gmrchk, line 25: The CSS escape seems to be very much supported. Do we need the fallback at all?

src/index.js
```diff
@@ -92,6 +92,7 @@ export default class Swup {
 		this.updateTransition = updateTransition;
 		this.getAnimationPromises = getAnimationPromises;
 		this.getPageData = getPageData;
+		this.getAnchorElement = getAnchorElement;
```
gmrchk, line 95: The naming doesn't make much sense, but I always used to separate modules vs. helpers as being available on Swup instance vs. not being available on Swup. I agree that this helper should be on Swup, but it's a bit breaking the convention this way, making things inconsistent. 
Should we move it under modules? Maybe even rename modules to something more fitting?

gmrchk (approved):

src/utils/index.js
daun: Happy to leave it out. Since global support is at 93% of devices, we should probably still leave it in a conditional. Worst case, nothing is escaped in pre-Chrome Edge.

src/index.js
daun: Let's move it under modules and think of a better naming strategy later :)

src/helpers/getAnchorElement.js
```diff
@@ -0,0 +1,19 @@
+import { escapeCssIdentifier } from '../utils';
+
+const getAnchorElement = (hash = '') => {
+	if (!hash) {
+		return null;
+	}
```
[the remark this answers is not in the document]
daun: Probably not, I added it out of habit since it adds code completion in the IDE.

src/helpers/getAnchorElement.js
```diff
@@ -0,0 +1,19 @@
...
+
+	if (hash.charAt(0) === '#') {
+		hash = hash.substring(1);
+	}
+
+	hash = decodeURIComponent(hash);
+	hash = escapeCssIdentifier(hash);
```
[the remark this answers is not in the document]
daun: `decodeURIComponent` decodes percentage-encoded data like `%20` back to normal ASCII.

`escapeCssIdentifier` escapes special characters for use in a CSS selector. Without that, IDs with colons and unicode characters throw an error in `querySelector`.

daun:
Thanks for the review. I'll work in the suggested changes and ping you again.

gmrchk (approved):

src/modules/getAnchorElement.js
```diff
@@ -0,0 +1,19 @@
...
+
+	if (hash.charAt(0) === '#') {
+		hash = hash.substring(1);
+	}
+
+	hash = decodeURIComponent(hash);
+	hash = escapeCssIdentifier(hash);
+
+	// https://html.spec.whatwg.org/#find-a-potential-indicated-element
+	return document.querySelector(`#${hash}, a[name='${hash}']`);
```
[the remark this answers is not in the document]
daun: Good catch, you're right, I'll fix it.

gmrchk (dismissed):

gmrchk (approved):

PR #448, opened 2022-07-20
Normalize cache URLs

### LINKED ISSUE #379: Document that loadPage specifically doesn't accept full URLs.
doing `swup.loadPage({url: window.location.toString()})` is problematic, at least when combined with `SwupBodyClassPlugin`, given setup like:
```
const swup = new Swup({
                cache: false,
                plugins: [
                    new window.SwupDebugPlugin(),
                    new window.SwupBodyClassPlugin(),
                ]);
```

you can end up with the following trace in your console:
```
TypeError: Cannot read properties of undefined (reading 'pageClass')
    at SwupBodyClassPlugin.js:174
    at swup.js:1167
    at Array.forEach (<anonymous>)
    at Swup.triggerEvent [as _triggerEvent] (swup.js:1165)
    at Swup.DebugPlugin._this.triggerEvent (SwupDebugPlugin.js:156)
    at Swup.renderPage (swup.js:1114)
    at swup.js:758
```
this is because `SwupBodyClassPlugin` does `var page = _this2.swup.cache.getCurrentPage()` which will return `undefined`

It's undefined because because `getCurrentPage` only uses `window.location.pathname + window.location.search` but `cacheUrl` just persists whatever is given as the URL, assuming `getPageData` didn't return null.

The simple solution is to document that `loadPage` takes hostname relative URLs (path onwards only), **but** 
ideally the situation should be normalised so that the part which looks in the cache is always consistent, such that passing a full URL to loadPage either trims off the extraneous prelude, or `getCurrentPage` takes the whole location into account and all cache keys are always based on the full URL...

daun:
Fix #379 by normalizing URLs before caching pages.


cypress/e2e/main.spec.js
```diff
@@ -148,9 +148,37 @@ context('Window', () => {
...
         });
     });
+
+    it('should update the body class', () => {
+        cy.window().then(window => {
+            window.swup.loadPage({ url: '/page2/' });
+            cy.get('body').should('have.class', 'body2');
+            cy.get('body').should('not.have.class', 'body1');
+        });
+    });
```
gmrchk, line 163: This seems like plugin functionality being tested. I think we could use more of that to ensure the plugins are working correctly after updating Swup core itself. Should we start separating those tests into different places?

gmrchk (commented):
Do you think it would make more sense to introduce some kind of Page object constructor that would already have a URL normalizer used within, like [here](https://github.com/swup/swup/blob/master/src/index.js#L123). It seems like doing it in Cache is too late, but maybe the idea is only to adjust the caching keys, not the rest of the functionality, in which case this sounds better. WDYT?

cypress/e2e/main.spec.js
daun: Good point. Seeing how we're already testing other plugin functionality here (scroll plugin), I guess we can leave this in for now and tackle the plugin tests in a future PR. Moving the tests to the plugin repos could get a bit messy unless we create a new support repo `@swup/cypress-config` or something. Monorepo could be a solution, too. I'll start a discussion in Slack to gather ideas.

daun:
> Do you think it would make more sense to introduce some kind of Page object constructor that would already have a URL normalizer used within, like [here](https://github.com/swup/swup/blob/master/src/index.js#L123). It seems like doing it in Cache is too late, but maybe the idea is only to adjust the caching keys, not the rest of the functionality, in which case this sounds better. WDYT?

For the moment, the idea was to adjust the caching keys. I can definitely imagine extending the Link object and including that in a Page object, but it's probably a slightly bigger change and currently I can't think of a specific problem we'd be solving, so I'd suggest tackling that when we see the need arise, no?

daun:
(Rebased on current master branch)

gmrchk:
I think the rebase didn't go too well, at least based on the [changes tab](https://github.com/swup/swup/pull/448/files) where there are plenty of changes marked from the master. Could you double-check?

daun:
Crap. I'll see if I can restore it.

daun:
@gmrchk I've force-pushed a rebased branch, should be good to review now.

src/modules/Cache.js
```diff
@@ -1,10 +1,13 @@
+import { getCurrentUrl, normalizeUrl } from './../helpers';
```
gmrchk, line 1:
```suggestion
import { getCurrentUrl, normalizeUrl } from '../helpers';
```

gmrchk (approved):

daun:
@gmrchk I think you'll need to review again after I commited your suggestion. Strange.

daun:
@gmrchk Good to review.

gmrchk (approved):

PR #449, opened 2022-07-20
Add error messages about transition containers

### LINKED ISSUE #397: Cannot set properties of null (setting 'responseURL')
So I was using Swup and for some strange reason it decided to stop working, I used the exact thing that the Demo page told me but anything works, on my `<html>` tag, doesn't appear the `class="swup-enabled"` and if I enter a link the page animates but the content doesn't appear until I refresh

**Here is what the console told me**:
`Cannot set properties of null (setting 'responseURL')`

This is the js file
```javascript
const swup = Swup();
```
Can someone help me please?

daun:
If no container was found on the page, the resulting console error is rather confusing: `Cannot set properties of null (setting 'responseURL')`. This PR adds two new console errors if a) no container was found or b) the number of containers between pages doesn't match.

Fixes #397 and similar future support requests by providing helpful error messages.

Includes some refactoring and adds a new test for `data-swup` attributes on containers.


gmrchk (approved):
I can't believe I didn't do this sooner. A super common mistake that is often ruining the first moments with Swup.

Issue #444, opened 2022-07-14
Swup breaks when animation-class and directory named the same


tobisalad:
I just spent several hours trying to find out why my very minimalistic swup-setup didn't work until I finally found what I believe is a very weird and apparently unknown bug: 
**When the directory your files are in is named the same as your animation-class then swup breaks.**
(by break I mean it fades out but never gets to the AnimationOutDone-Event and thus never transitions to another page)

In my case I had my files in a webroot-subdirectory `/transition/`, thus my links were of the form `href="/transition/page.html"`, and I was working with the default animation-class `transition-*`. I confirmed this behavior by renaming both my directory and animation-class to something else and it was still broken, as soon as the two are not named the same it would work again.

daun:
That's definitely a bug. Swup adds `to-*` classes to the `html` tag to allow switching animations between different subpages. If the URL of the next page contains the word _transition_, it will find the class it just added itself and wait for a transition on the `html` tag — which will never happen.

Setting a custom selector for the animations should solve this however. If you add `html` before the actual selector, swup will only look for transitioned elements inside the `html` tag.

```js
const swup = new Swup({
  animationSelector: 'html [class*="transition-"]'
})
```

daun:
Fixed in [2.0.17](https://github.com/swup/swup/releases/tag/2.0.17)

daun:
@tobisalad The workaround above will stop working in the new version as all selectors will now be scoped to the `body`.

tobisalad:
Thanks! :-)

PR #455, opened 2022-07-26
Resolve immediately if no transitions defined

daun:
This change will help swup deal with a few common malconfigured cases:

- no animated elements were found → immediately resolve animation promise
- animated element has no CSS transitions → immediately resolve animation promise

The main goals here are:

- provide a happy path even if there's no animated element, defaulting to a `0s` timeout  
- reduce support burden by adding helpful error messages for the most common mistake

This will also fix:

- issue #444 by querying for animated elements below the `html` element, making sure the `from-*` classes added by swup won't be used as animation promises if the URL contains the word `/transition/`.


src/helpers/transitionEnd.js
```diff
@@ -1,20 +1,9 @@
...
-	};
-
-	for (let name in transEndEventNames) {
-		if (el.style[name] !== undefined) {
-			return transEndEventNames[name];
-		}
+	if (window.ontransitionend === undefined && window.onwebkittransitionend !== undefined) {
+		return 'webkitTransitionEnd'
+	} else {
+		return 'transitionend'
```
gmrchk, line 5: Not sure why is this change needed. Are the values different nowadays?

src/helpers/transitionProperty.js
```diff
@@ -0,0 +1,9 @@
+const transitionProperty = () => {
+	if (window.ontransitionend === undefined && window.onwebkittransitionend !== undefined) {
+		return 'WebkitTransition'
+	} else {
+		return 'transition'
+	}
```
gmrchk, line 6: Looks like duplicate code. Code we reuse together with `transitionEnd` function?

src/modules/getAnimationPromises.js
```diff
@@ -1,10 +1,23 @@
...
+	if (!animatedElements.length) {
+		console.error(`No animated elements found by selector ${this.options.animationSelector}`);
+		return [Promise.resolve()];
+	}
+
 	animatedElements.forEach((element) => {
+		const transitionDuration = window.getComputedStyle(element)[`${transitionProperty()}Duration`];
+		// Resolve immediately if no transition defined
+		if (!transitionDuration || transitionDuration == '0s') {
+			console.error(`No CSS transition defined for element of selector ${this.options.animationSelector}`);
```
gmrchk, line 17:
```suggestion
			console.error(`No CSS transition duration defined for element of selector ${this.options.animationSelector}`);
```

src/helpers/transitionEnd.js
daun: I thought the `opera` part looked dated, then just looked at how Vue currently handles transitions as it's baked into the core and works really well, and that's how they do it 🤠 We could revert this part to how it's done currently but this one is less code. However the opera part is not necessary anymore since they switched to Chromium as well. Want to revert it?

src/helpers/transitionProperty.js
daun: We could turn those two into one function that returns both? Or set a Webkit flag. Personally, I have a rule only to refactor and DRY for 3 or more occurrences of similar code so this one has flown under the radar 🤡

src/helpers/transitionEnd.js
gmrchk: Nah, just curious, let's roll with it!

gmrchk (approved):

Issue #187, opened 2019-08-16
Buggy when hitting back button during page transition


Yvee1:
When one hits the back arrow during a page transition, the result is buggy. On the swup documentation site for example, I can get an empty page with trailing classes on the `html` element.

![screenshot of emtpy page](https://user-images.githubusercontent.com/30909373/63176860-01fbc900-c047-11e9-8648-f2289c42ff82.png)

On the decathlon usage example page, when I hit back during the 'out' transition I do get a non-empty page with no trailing classes, but it is the incorrect page.

![screenshot of page](https://user-images.githubusercontent.com/30909373/63177029-5acb6180-c047-11e9-9735-45a65cf573b1.png)

You can see the url indicates the homepage, but another page is showing.

The problem is the biggest during an out-transition, but during an in-transition on my own page for example there are some trailing classes (`is-changing to-mehtml is-rendering` to be exact) though they don't cause any problems I think.

I do enjoy the library though, really easy to use!

gmrchk:
Hi @Yvee1, thank you for reporting, this has been on my mind for a while. 

Question is, what's the correct approach here. Block back/forward when the transition is in the process? Stop animation and replace the content right away? Some other approach?

I think I would go for a second option.

Yvee1:
I think blocking the buttons is indeed not a good idea. 

I would suggest that when the back button is clicked during an out-transition, all swup classes are removed from the html element hence stopping (or reversing I guess?) the animation, and the url is changed back to what it was before. If the back button is clicked during an in-transition it works fine how it is currently, but maybe all the swup classes on the html element can be removed.

I guess the same would hold for the forward button.

idflood:
Here is a partial fix (when clicking quickly on multiple links):

```javascript
swup.on('clickLink', () => {
  document.documentElement.classList.remove('is-leaving');
  document.documentElement.classList.remove('is-animating');
});
```

hirasso:
This has been addressed in [2.0.18](https://github.com/swup/swup/releases/tag/2.0.18)

PR #463, opened 2022-08-09
Handle rapid navigation

hirasso:
Fixes the bug described in #187 .

This attempt addresses the described bug by doing two things:

1. Don't execute `renderPage` if `page.url` doesn't match `getCurrentUrl()`:

```js
const isCurrentPage = this.getCurrentUrl() === page.url;
if( !isCurrentPage ) return;
```
2. Remove all swup-related html-classes on popstate, if `animateHistoryBrowsing` is `false`:
```js
if( !this.options.animateHistoryBrowsing ) {
  document.documentElement.classList.remove('is-animating');
  cleanupAnimationClasses();
}
```
I tried to run cypress on this, but I can't get it to work / I have never used it. Manual testing looked fine, though.


daun:
We're probably over-engineering this. If we set an `isAnimating` flag on the instance and force a normal browser load (`window.location = *`) for all links that are clicked while animating, we get rid of all the edge cases and keep complexity to a minimum. What do you think?

hirasso:
This fix doesn't add a lot of complexity and it works fine for me. Actually, I was quite surprised that the check for the still-current-url in `renderPage` wasn't there, already.

Have you tested it?

hirasso:
I uploaded a little test site to my server:

https://test.rassohilber.com/swup-handle-rapid-navigation/test/

The transition is extra long there, so that you have plenty of time to hit the browser back/forward buttons ;)

I made the `swup` instance available in the `window` object, so you can change `animateHistoryBrowsing` at runtime: 
```js
swup.options.animateHistoryBrowsing = true
```

daun:
It's working surprisingly well! And you're right, it's not that complex. All the whitespace has changed from tabs to spaces, that's why it looks more complex than it is 🌝 Maybe you can fix that?

One edge case to keep in mind: 301/302 redirects. Can we test that as well?

hirasso:
Haha sorry, the whitespace problem again 🙈 ...I'll try to fix it. 

Regarding 301/302 redirects: We don't want this to happen, either, if the requested URL has changed. Hence the early conditional return in `renderPage`.

daun:
So what happens if I click on /info/ and the server redirects to /info/about/? Does the current behavior change or will it abort at some point?

hirasso:
If you wait for the page to load, the current behavior stays like is. If you navigate somewhere else before the page is loaded, any detected redirect is being ignored. We don't need the result right now anymore, after all.

daun:
What I meant is: does that new `currentUrl()` vs. `page.url` check also work in case of redirects?

hirasso:
Yes, `page.url` is not being altered by a redirected server response, so it should work. Haven't tested it, though.

daun:
Great, I'll check it out locally.

hirasso:
I managed to get prettier to work. Just ran it on the two files I modified, it cleaned up my code, but also quite a bit of code that wasn't part of this PR.

```shell
npx prettier --write ./src/modules/renderPage.js
npx prettier --write ./src/index.js
```
We should definitely think about introducing some automated invocation of `prettier` IMO.

daun:
Can confirm this is working nicely. Tests are passing and it works for normal clicks, popstate and redirects. Great work 🥯

daun:
Can you create a branch on the swup repo itself, not your fork? I think that's why the CI build isn't happening.

daun (approved):

hirasso:
Closing this PR in favor of https://github.com/swup/swup/pull/464

PR #467, opened 2022-08-10
Fix: Handle switching contexts

hirasso:
People might set up Swup differently or not at all in some areas of their websites:

```js
// Initially the containers option could be empty and filled-up based on context
const options = {
  containers: []
};
// This would resolve to any element with an ID that starts with "swup-"
const primaryContainer = document.querySelector('[id^="swup-"]');
if( primaryContainer ) options.containers.push(`#${primaryContainer.id}`);
// Also, there could be different amounts of containers in different contexts:
const secondaryContainer = document.querySelector('[id="secondary-swup-container"]');
if( secondaryContainer ) options.containers.push(`#${secondaryContainer.id}`);

const swup = new Swup(options);
```
The above example would create 3 contexts:
- one without swup being active, at all (containers: [])
- one with one container that has an id attribute that starts with `swup-`
- one with two containers, where the second has the id `secondary-swup-container`

If navigating between areas that have different or none swup containers, swup would never resolve. Now it does a hard page reload if `page.blocks` is empty.

Found this while investigating #260

Maybe it would be worth mentioning somewhere in the docs, that the container ids should always be completely different for different contexts?


hirasso:
Created a demo here:

https://test.rassohilber.com/swup-handle-switching-contexts/test/

daun:
Looking good! I agree that console.warn is enough if we're actually handling the issue.

daun:
If you build the dist files as well, it's ready to merge.

hirasso:
I keep forgetting building these darn dist files 😄

daun (approved):

PR #471, opened 2022-08-11
Define custom global instance for e2e tests

daun:
Prepare for upcoming [debug plugin updates](https://github.com/swup/debug-plugin/pull/13):

- no longer rely on implicit `window.swup` created by debug plugin
- instead, define our own `window._swup` instance for e2e testing

### DIFF
--- a/cypress/e2e/main.spec.js
+++ b/cypress/e2e/main.spec.js
@@ -1,5 +1,7 @@
 /// <reference types="Cypress" />
 
+// window._swup holds the swup instance
+
 context('Window', () => {
     beforeEach(() => {
         cy.visit('/page1/');
@@ -33,7 +35,7 @@ context('Window', () => {
     it('should trigger a custom click event', () => {
         let triggered = false;
         cy.window().then(window => {
-            window.swup.on('clickLink', () => triggered = true);
+            window._swup.on('clickLink', () => triggered = true);
         });
         cy.triggerClickOnLink('/page2/');
         cy.window().should(() => {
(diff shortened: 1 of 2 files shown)


daun:
@hirasso By the way, to run the tests locally, you'll need to remove the `--record` flag from the `test:run` npm script. Something we'll need to change and/or document. They're currently passing for this PR, however.

hirasso:
Ah, that's good to know, thanks! Already wanted to create an issue to ask just that :)

hirasso:
Just ran `npm run test`. It fails, outputting stuff that I don't understand:
Screen Shot 2022-08-12 at 10 49 08

daun:
@hirasso That's a 404, the server doesn't seem to pick up or serve the html files.

daun:
Do you have a different server running at that port already?

hirasso:
Ok, found the issue. The port `8080` is occupied on my machine, apparently, and `http-server` took the next free port:

Screen Shot 2022-08-12 at 10 55 57

Changing the port in `cypress.config.js` solved it for me. Also, there is a hardcoded server in [main.spwc.js](https://github.com/swup/swup/blob/ee1631d99b8b329280eb8cfcc7353150295d6be1/cypress/e2e/main.spec.js#L192), to test for absolute URLs. Had to change that, as well.

Now the tests are passing! If there was a way to handle this kind of scenario (occupied port) automatically, that would be awesome.

daun:
The easiest solution I can thinknof is moving to a different random port altogether, like 8274 or something.

hirasso (approved):

hirasso:
Do you want to take care of changing the port or should I?

daun:
@hirasso Maybe you do it, it might give you some more familiarity with the test setup in general.

hirasso:
A question: The test generates a video under /cypress/videos/main.spec.js.mp4. Should this really checked-in, as well?

daun:
@hirasso Haven't thought about it, I just discard all changes to it whenever it fails. We could probably gitignore it, yeah. Let's ask Georgy at some point.

PR #472, opened 2022-08-12
Change test server port from `8080` to `8274`

hirasso:


cypress/e2e/main.spec.js
```diff
@@ -189,7 +189,7 @@ context('Window', () => {
 
     it('should cache pages from absolute URLs', () => {
         cy.window().then(window => {
-            window._swup.loadPage({ url: 'http://localhost:8080/page2/' });
+            window._swup.loadPage({ url: 'http://localhost:8274/page2/' });
```
daun, line 192: Could we somehow get this as a variable from cypress.config.js? Might come in handy if we keep referencing absolute URLs in tests. Or maybe that's overkill, seeing how it's hardcoded all over the place anyway? Your judgement :)

hirasso: Let's keep it dumb for now, ok? Changing the port is a simple search-replace, after all :)

daun: Okay :)

daun (approved):

PR #477, opened 2022-08-16
Remove issue templates and funding information

daun:
I've moved and improved on the issue and pull request templates and funding info; they're now in the [swup/.github](https://github.com/swup/.github) repo. Those templates are applied on all other repos as well, i.e. we only ever have to edit them in one place. It also means we should delete the local ones so that they are inherited correctly.


hirasso (approved):
Wow, how do you know this stuff?! 🔥

daun:
Interestingly, it doesn't yet apply the issue template...

daun:
Okay, got it working.

PR #478, opened 2022-08-16
Fix cache init

### LINKED ISSUE #475: Delay caching first page until after DOM load event
When a first page-load has `<style>` tags in a container used by Swup, it will not add them to the cache. Navigating back will result in the page not having the `<style>` tags that it initially existed.

This issue does not present itself with subsequent page-loads, loaded by mouse-over, for example. It that case, the tags get added to the cache properly.

### Workaround
For now, you can turn off the cache:
```js
const options = {
  cache: true
}
```

This isn't ideal (the caching Swup does is the reason I'm using it, and it rocks when it works), but this will prevent any lost styles. The issue is especially pungent on Wordpress 5.9+ sites that insert dynamic `<style>` blocks.

daun:
- Check and wait for `DOMContentLoaded` event before caching initial page
- Fixes #475


hirasso:
I'm not sure if waiting for `DOMContentLoaded` for the first cache entry wouldn't introduce many new problems. Inline `<script></script>` tags would have been executed already at this point, for example. The only clean way to retrieve a guaranteed un-modified version of the current page would be by doing a `fetch` to the current page and storing the response of this to the cache. What do you think?

daun:
How about not caching the first page at all? That's how barba seems to do it, I assume after they've seen one too many related issues...

daun:
Running a fetch is a bit aggressive I think, clogging the network without asking. Then it's a question of waiting for the network to become idle, and that's not really support across most browsers.

hirasso:
As a hotfix for above bug we could just disable the cache for the first page, I suppose. If we find out later, that we actually need it, we can still add the fetch. Support for `requestIdleCallback` isn't that bad anymore, by the way: https://caniuse.com/requestidlecallback

daun:
Alright, I'll update the PR.

The problem with requestIdleCallback is that it has no relation to network idling, only cpu idling. So the page might be idle but it's busy loading other stuff in the background. Maybe we should disable it by default and make the fetch logic part of the preload plugin, as installing that signals intent to have that kind of thing happen.

daun:
Closed in favor of [swup#480](https://github.com/swup/swup/pull/480) and [preload-plugin#29](https://github.com/swup/preload-plugin/pull/29).

hirasso:
Sounds good to me!

Issue #29, opened 2018-08-16
Yarn Package missing /dist


MikeHarrison:
When I add Swup through Yarn (`yarn add swup`), the package I get is missing the `/dist` folder. The folder tree I get is like:

```
swup
|- .idea
|- lib
|- plugins
```
And there is no swup.js present. Wierd one! Am I downloading the wrong package or is this an error?

gmrchk:
I'm going to explain some concepts that are used in swup file structure, so we're on the same page. 

When getting the package through Yarn or NPM, you get a package published on NPM. 
The `package.json` file includes a line `"main": "lib/index.js",`, which tells NPM that this is the root file, so when you do `import Swup from 'swup'`, this is the file you're getting. Actually, this is only a root file of swup which imports all other parts it needs from *lib* folder, so basically what you are loading is the whole *lib* folder. *lib* folder is a copy of *src* folder and is transpiled by babel, so people can load it from NPM and doesn't have to transpile it again. 

Plugins folder is in the root so the plugins can be loaded like this `swup/plugins/pluginName`. This folder is in the root only because NPM does not allow to define root folder, which in this case, would be `lib` and loading plugins as `swup/lib/plugins/pluginName` just seems too ugly. 

`dist` folder is already bundled code ready for browser, that's why it can be included in a script tag. 

NPM similarly to GIT has a [.npmignore](https://github.com/gmrchk/swup/blob/master/.npmignore) file which, in this case, includes dist folder, so it is not published in NPM, as it is not necessary because all you need is in lib folder which is accessed when you do the `import Swup from 'swup'`.

---

Now, when people use a package manager like Yarn or NPM to get packages, I assume they are using some bundler to bundle it all together, and not include the bundled file in script tag from `node_modules` folder (or at least I never met anyone who does). That's why it's not included in the NPM package and only present here on Github where people can download it if needed. Do you load scripts from `node_modules` folder via script tag? There is no problem to add it. I just always thought it's unnecessary and wouldn't be used. 

PS:
`.idea` is folder used by my IDE and shouldn't be in the package. I have deleted it. Thank you for pointing it out.

MikeHarrison:
Hi,

Thanks for taking the time to explain, I appreciate it :)

Perhaps my workflow is a little odd! I have my `node_modules` folder, and use Yarn to add / upgrade / delete packages within it.

[...]

PR #480, opened 2022-08-16
Disable caching of initial page

### LINKED ISSUE #475: Delay caching first page until after DOM load event

daun:
- Disable caching of initial page state to avoid modified dom being cached
- Logic is moved to preload plugin: see [#29](https://github.com/swup/preload-plugin/pull/29)
- Replaces #478 
- Fixes #475
- Waiting on #484 to fix test suite


hirasso (dismissed):

hirasso:
Hmm, something off with the tests here. At first, I thought it was related to the cache tests, but it doesn't seem so. The error message I am getting also doesn't help a lot:
```shell
1) "before each" hook for "should transition forward to page 2 on popstate"
```

daun:
@hirasso Disabling the cache seems to break the tests as well, even without the changes in this PR. I'll look into it.

daun:
Phew. This one isn't easy. I'd love to fix this for the next release but there's no rush I guess.

daun:
@hirasso With the recent updates to the test suite, the tests are passing again. Please re-approve whenever you get the chance.

hirasso (approved):

PR #484, opened 2022-08-21
Fix history browsing test suite

daun:
With the `cache` option set to `false`, the tests would reliably fail due to an obscure CORS issue with Cypress.

This PR does a few things to address the situation and improve testing in general:

- Disable Chrome web security during test runs (the actual fix)
- Resolve XHR promises with the page data instead of getting it from the cache, thus removing hard-to-understand dependencies in the code
- Allow for sub-pixel precision in scroll-position tests, fixing tests in Firefox
- Add an npm script for headed test runs, allowing inspection of the Cypress instance and the test site between test runs
- Update Cypress


hirasso (approved):
Tests are passing, great! 👏

hirasso:
One thing I keep getting in the end:
```shell
[2022-08-22T10:53:24.780Z]  "GET /__coverage__" Error (404): "Not found"
```

daun:
Ah yes, the coverage has never worked correctly; something we can fix later.

Issue #490, opened 2022-08-21
New bundler


gmrchk:
We've talked about replacing the bundler to allow some easier work. The webpack setup is also outdated in basically all repositories including Swup core and plugins. 

We've agreed that we would try:
- [x] @gmrchk to create PR with updated webpack config in the Swup core. 
- [x] @daun to create PR with replaced bundler providing some more straightforward setup.

gmrchk:
This would ideally also solve #487.

daun:
@gmrchk I've created two draft PRs for both versions: #492 and #493. Feel free to check out both locally and test them. We can also move the discussion to slack.

daun:
Waiting for release in 3.0

PR #493, opened 2022-08-21
Bundler: integrate microbundle

daun:
Option 2 of the bundler revamp in #490 — integrate microbundle.

See [Writing a small module](https://macwright.com/2018/11/26/small-modules.html) for some of the assumptions and ideas.

- Remove webpack & entrypoint file
- Remove babel (already included in microbundle)
- Remove `lib` and `dist` folder; now only `dist` is generated and doesn't have to be tracked in git
- Create all sorts of optimized bundles: CommonJS, UMD, ES module and module for modern browsers
- Simple maintenance: just upgrade microbundle
- TypeScript support is built in, so no extra setup required in case we decide to migrate the codebase to TS

Things left to determine/discuss:

- Currently a breaking change since the `lib` folder is not necessary anymore and some plugins directly import from the nested `lib/helpers/` folder. My suggestion here:
  - Re-export all helpers as named exports from the entrypoint so that they can be imported in one place:   
    `import Swup, { getPageData, getCurrentUrl } from 'swup'`
    Then update all plugins to use those new named exports.
  - To not break systems importing from the current `lib/helpers`, we can create an export map ([example](https://docs.skypack.dev/package-authors/package-checks#export-map)), recreating that folder for bundlers only. The imports would keep working; however we should deprecate the old paths in the release notes and remove them in the next major version.


hirasso:
If we choose this approach (I like it!), could we also add a `watch` npm script?

daun:
@hirasso Sure, `microbundle watch` would do just that. I can add another script for that.

daun:
I've add a `dev` script with watch option enabled, merged the current master und updated dependencies.

gmrchk:
What can I say, this looks great. Massive simplification. 🥇 

I would like to test it out with webpack builds, loading as a module in the browser, and whatnot, just to make sure nothing is broken, but I don't see why we would give it a go!

Nice thinking about the backward compatibility with plugins. Didn't know you can do that with the `exports` field. 👍 

One thing to note - people often rely on the dist folder to download the project from GH and copy it from there. That's why it's there in the first place... 
I never liked it and it's pain for building the files with each release also, but we should document an alternative solution for the same downloading of the UMD bundle.

daun:
> I would like to test it out with webpack builds, loading as a module in the browser, and whatnot, just to make sure nothing is broken, but I don't see why we would give it a go!

That's my plan, use it in a few projects on webpack 4, webpack 5 and via script tag to see if it works everywhere.

> we should document an alternative solution for the same downloading of the UMD bundle

We can point people to the unpkg page which has a list of downloadable files: [unpkg.com/swup/](https://unpkg.com/swup/). The skypack page has a usage example on top for importing from a CDN: [skypack.dev/view/swup](https://www.skypack.dev/view/swup).

PR #498, opened 2022-08-30
Support CSS keyframes and multiple transition durations

daun:
Feature addition to allow multiple transition durations as well as animation keyframes.

**Things that currently don't work**

- Combining multiple transitions with different durations and having both finish: swup will only wait for the first transition to finish before replacing the content
- Using `animation` keyframes for page transitions: swup only supports the `transition` property

**What this PR does to make those things work**

- Wait for all transitions to finish before replacing the content
- Detect and work with CSS animations as well, choosing whichever is longest

**Prior art**

Most of the new behavior is lifted straight from [Vue's transitions module](https://github.com/vuejs/core/blob/main/packages/runtime-dom/src/components/Transition.ts).  
It's working really well so far on a lot of production sites of mine.

**Questions**

- Is there any argument to be made for NOT waiting for all transitions to finish?
- Should we prefer transitions to animations? The PR currently defaults to whichever has the longest duration in case both are defined.

Notes

- Added tests for simple and complex transitions as well as animation keyframes
- Looks somewhat complex, but will increase bundle size by barely 0.05KB


hirasso:
While trying to test this I got compile time errors, probably related to the planned bundler update? They are all the same, for different imports:

```shell
ERROR in ./swup/src/index.js 18:0-36
Module not found: Error: Can't resolve './modules/Cache' in '/Users/rah/Sites/swup/wwwroot/general-tests/swup/src'
Did you mean 'Cache.js'?
BREAKING CHANGE: The request './modules/Cache' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.

ERROR in ./swup/src/index.js 19:0-42
Module not found: Error: Can't resolve './modules/loadPage' in '/Users/rah/Sites/swup/wwwroot/general-tests/swup/src'
Did you mean 'loadPage.js'?
BREAKING CHANGE: The request './modules/loadPage' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.

ERROR in ./swup/src/index.js 20:0-46
Module not found: Error: Can't resolve './modules/renderPage' in '/Users/rah/Sites/swup/wwwroot/general-tests/swup/src'
Did you mean 'renderPage.js'?
BREAKING CHANGE: The request './modules/renderPage' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
```

...and so on

daun:
@hirasso Can you share how you're bundling it? I'm wondering why it is directly including `swup/src/index.js` and not the build files from `dist`.

My webpack setup seems to not be that strict, although it probably should.

hirasso:
Ahh right, changing the import from `import Swup from '../../swup/src/index.js'` to `import Swup from '../../swup/'` fixed the errors. Probably due to the new `exports` entry in the `package.json`?

daun:
I see. The package is now declared `type: module` so all imports require a file extension. But the dist files are already bundled, so nothing to import there. For importing from `src`, I'll need to add the file extensions in the codebase.

daun:
Technically, you can tell webpack to not require fully resolved filenames for ESM imports.

hirasso:
Let's still change all imports to `.js`

hirasso:
There were still some unqualified imports in `./helpers/index.js` after fixing these, it compiles now for me.

hirasso:
...just tested this, for now only with two different `.transition-x` classes. While the out-animation works as expected (waits for the longer transition to complete before continuing), the transition-in events (`animationInDone`, `transitionEnd`) are being fired as soon as the first transition finishes. Shouldn't this also wait for the longest transition?

daun:
@hirasso Looks like you found a bug then 🤠 Can you share the composition/durstion of your transitions for me to test out?

hirasso:
Sure, tomorrow :)

hirasso:
I found some time already.  Here's my CSS for the transitions:

```css
.transition-short {
  transition-property: opacity;
  transition-duration: 200ms;
  opacity: 1;
}
html.is-animating .transition-short {
  opacity: 0;
}
.transition-long {
  transition-property: background;
  transition-duration: 5000ms;
  background: white;
}
html.is-animating .transition-long {
  background: red;
}
```
I am testing the new functionality like this:
```js
const swup = new Swup();
swup.on('animationOutStart', e => console.time('animationOut'));
swup.on('animationOutDone', e => console.timeEnd('animationOut')); // Takes 5003.696044921875 ms
swup.on('animationInStart', e => console.time('animationIn'));
swup.on('animationInDone', e => console.timeEnd('animationIn')); // Takes 197.341064453125 ms
```
My very reduced HTML:
```html
<body>
  <h1 class="transition-long">Swup General Tests</h1>
  <main id="swup" class="transition-short">...</main>
</body>
```

daun:
The problem seems to occur if you don't add both containers to swup. This helps:

```js
const swup = new Swup({
  containers: ['h1', 'main']
});
```

Now the question is: should this be supported? Not sure if this is currently something that'd work.

daun:
I've pushed a fix. Basically, the element not being in swup's containers prevented it from being replaced by `outerHTML`, and the `transitionend` event would 'bleed' from the out-phase into the in-phase. The solution is just checking for elapsed time to discard outdated events.

daun:
BTW: Your example with two elements would already work as expected in the current master version, i.e. waiting for the longer transition of both elements to finish.

The *actual* use case this is supposed to enable is having multiple transition timings on one single element. The example below would resolve after 400ms in the current version and after 2000ms with this PR merged.

```css
.transition-short-and-long {
  transition: background 300ms 100ms, color 1500ms 500ms;
}
```

hirasso:
Ah sorry, I hadn't really understood the problem this PR is trying to solve, yet. Just tested it with both scenarios – different transition durations for different properties as well as with '@keyframes` animations – it works well! 🍰

...I got confused because of the new paragraph we added to the docs a while back: 

> When a user navigates your website, Swup waits for all elements on the current page that contain the class transition-[something] to finish their animation, before switching the container contents and animating your page back in. We recommend that you set the transition-[something] class on only one element for each page. All other elements you want to animate during page transitions should be animated independently (without using a separate transition-[something] class for each of them).

I actually thought this was about that. Could you remind me why we are recommending this again?

daun:
@hirasso I guess that sentence is no longer necessary, then. It wasn't technically required and it's still a good way to reason about things, but might confuse a few people if it's not strictly necessary for most sites.

daun:
@hirasso I've added tests for all types of transition durations, combinations and delays. Looking good. I'd say ready to merge here.

hirasso:
Leet's go! 🍹

PR #499, opened 2022-08-31
Fragments

daun:
## Features

- Replace page fragments nested in containers, without replacing the whole container
- Disable animations for visits with fragments

## Questions

- Should fragment visits be animated at all for a first feature release?
  - Can this be a later release? Or better to nail down API now?
- Which events should be fired for fragment visits?
  - Especially re: integration with scroll/progress/a11y plugin that rely on `contentReplaced`
  - If firing `contentReplaced` has too many side effects, we need a `fragmentReplaced` event, but then some of the plugins will need a rewrite as well

## Open issues

- [x] Create playground
- [ ] Find solution for animating fragment transitions: (classes on fragment? on html tag?)
- [ ] Thoroughly test popstate visits
- [ ] Write Cypress tests

## API

**Defining fragments**

```html
<main id="swup"> <!-- Normal container -->

  <nav>

    <!-- Will trigger a fragment visit, replacing both fragments -->
    <a href="/" data-swup-to-fragment>Replace fragments</a>

    <!-- Will trigger a fragment visit, but only replacing the `list` fragment -->
    <a href="/" data-swup-to-fragment="list">Replace example fragment</a>

    <!-- Fragment doesn't exist, will trigger a normal visit -->
    <a href="/" data-swup-to-fragment="missing">Try to replace missing fragment</a>

  </nav>

  <div data-swup-fragment-container="list"> <!-- Fragment -->
    Consectetur adipisicing elit
  </div>

  <div data-swup-fragment-container="pagination"> <!-- Fragment -->
    Page 1 of 2
  </div>

</main>
```

The fragment target attribute can also be defined on a parent, confining all child links to that fragment.

```html
<!-- All links in this nav will now force a fragment visit -->
<nav data-swup-to-fragment>

  <!-- [data-swup-to-fragment] not required anymore -->
  <a href="/">Replace fragments</a>

  <!-- Use special fragment name `_top` to force a normal visit -->
  <a href="/" data-swup-to-fragment="_top">Replace whole page</a>

</nav>
```


daun:
@hirasso Feel free to check out the current state of the branch. Looking forward to your input on the API, naming and handling of edge cases.

hirasso:
@daun exciting!! Looking forward to taking this for a ride :)

hirasso:
@daun could you merge the qualified imports into this branch? This exceeds my git knowlegde 😅

daun:
@hirasso I've merged the `next` branch into this one. It does compile, but let me know if something's funny.

src/modules/leavePage.js
```diff
@@ -0,0 +1,28 @@
...
+	this.triggerEvent('animationOutStart');
+
+	// handle classes
+	document.documentElement.classList.add('is-changing');
+	document.documentElement.classList.add('is-leaving');
+	document.documentElement.classList.add('is-animating');
+	if (popstate) {
+		document.documentElement.classList.add('is-popstate');
+	}
+	document.documentElement.classList.add(`to-${classify(data.url)}`);
```
gmrchk, line 18: Maybe this could have some util function as well, similar to the `cleanupAnimationClasses` introduced? I'm thinking something like, but maybe there is better options too. 

```javascript
addAnimationClasses('is-changing', 'is-leaving', 'is-animating', `to-${classify(data.url)}`);
if (popstate) addAnimationClasses('is-popstate');
```

src/modules/loadPage.js
```diff
@@ -1,98 +1,68 @@
-import { classify, createHistoryRecord, fetch } from '../helpers.js';
+import { classify, createHistoryRecord, updateHistoryRecord, fetch, getCurrentUrl } from '../helpers.js';
 
-const loadPage = function(data, popstate) {
-	// create array for storing animation promises
-	let animationPromises = [],
-		xhrPromise;
-	const animateOut = () => {
-		this.triggerEvent('animationOutStart');
+const loadPage = function(data, popstate = false) {
```
gmrchk, line 3: Would it make sense to rewrite this to an async function, so we could handle calling and awaiting both `leavePage` and `renderPage` inside of here? Maybe it could put all the handling of classes nicely into one place. 
Right now (historically) it seems like it could be unclear why are some classes added/removed at different times.

src/modules/renderPage.js
```diff
@@ -1,67 +1,103 @@
...
 	document.documentElement.classList.remove('is-leaving');
 
 	const isCurrentPage = this.getCurrentUrl() === page.url;
-	if (!isCurrentPage) return;
+	if (!isCurrentPage) {
+		console.log('not current page:', this.getCurrentUrl(), 'vs', page.url, page);
+		return;
+	}
+
+	const skipTransition = fragment || (popstate && !this.options.animateHistoryBrowsing);
```
gmrchk, line 13: I've seen this in a couple of places. Maybe also a good candidate for util function?

src/modules/renderPage.js
```diff
@@ -1,67 +1,103 @@
...
+				container.outerHTML = html;
+			}
+		});
+		this.triggerEvent('contentReplaced', popstate);
+		this.triggerEvent('fragmentsReplaced', popstate);
+	};
+
+	// Try fragments first, if not, then blocks
+	if (fragment) {
+		if (replaceFragments() === false) {
```
gmrchk, line 79: Can something other falsy be returned that it's checked specifically? This check is fine for sure too, but it might be prone to breaking in the future with a similar suggestion I have if there can be other falsy values for such function. 

```suggestion
		if (!replaceFragments()) {
```

src/modules/renderPage.js
```diff
@@ -1,67 +1,103 @@
...
+			console.log('Fragments failed, replacing blocks');
```
gmrchk, line 80: I think better to keep the terminology consistent.

```suggestion
			console.log('Fragments failed, replacing containers');
```

src/utils/index.js
```diff
@@ -14,6 +14,14 @@ export const queryAll = (selector, context = document) => {
 	return Array.from(context.querySelectorAll(selector));
 };
 
+export const nextTick = (callback) => {
+	requestAnimationFrame(() => {
+		requestAnimationFrame(() => {
+			callback();
+		});
+	});
+};
```
gmrchk, line 23: Love this! 

I was seeing it through the code and was preparing to comment `requestAnimationFrame` being a bit dangerous for such a case as things can manage to happen in one "tick" even if you don't want to, but I see you've got it covered with two. 😄

src/index.js
```diff
@@ -39,7 +42,9 @@ export default class Swup {
 			plugins: [],
 			skipPopStateHandling: function(event) {
 				return !(event.state && event.state.source === 'swup');
-			}
+			},
+			fragmentTargetAttr: 'data-swup-to-fragment',
+			fragmentContainerAttr: 'data-swup-fragment-container'
```
gmrchk, line 47: I would think one of these would end up being `data-swup-to-fragment`, probably the second one. I think we should avoid the word `container` when it comes to fragments, to prevent any confusion with the classic containers. 

I'm also thinking the options could probably go in one of the ways:  
1. Be a completely editable selector like the containers are, but then I get that we need to be able to save a value in those attributes as well. 
2. Be a fixed name.

Do you see some value in it being editable with the way the fragments API is set up? It seems like it just adds a bit of complexity in the setup instead of helping anything. Is there some situation where the attribute name would be better to be custom?

gmrchk (commented):
Some people might be overwriting some of the internal methods like `leavePage` or `renderPage` of Swup to do a custom handling, in which case things will break if any of the logic moved around.
I don't think we see it as a valid API usage, but still, should we consider communicating it in some way, or do you think minor bump enough?

src/modules/loadPage.js
daun: I'd love to rewrite everything with `async/await`, but that's one step too high on the complexity ladder for my taste. I'd prefer to get the functionality working and tackle async-await for the whole codebase as a separate refactor PR.

src/utils/index.js
daun: Yeah, learned that the hard way 🤠

src/modules/renderPage.js
daun: Sounds good!

src/modules/renderPage.js
daun: I have a rule to extract util functions if a line is repeated in 3 places or more. Is there a 3rd one I missed? I was counting 2.

src/modules/leavePage.js
daun: Good idea! The repetition of `document.documentElement` is a good pointer here for sure.

src/index.js
daun: I put them into the options for convenience while building the feature; I agree these probably shouldn't be editable and could get removed from the options. The advantage here is that the options object is basically available everywhere and we don't need to figure out how to pass the selector around.

daun:
> Some people might be overwriting some of the internal methods like `leavePage` or `renderPage` of Swup to do a custom handling, in which case things will break if any of the logic moved around. I don't think we see it as a valid API usage, but still, should we consider communicating it in some way, or do you think minor bump enough?

Valid point. Haven't thought about it too much, but we do have the luxury of introducing this as a breaking change and major version. We do however need to communicate it clearly, yes — ideally with a useful and detailed changelog and/or migration guide with warnings like these so that people with custom implementations don't go crazy.

hirasso:
The more I think about built-in fragment support, the more I get convinced that it might not be solvable. Handling all possible state changes through the browser's history navigation (backwards, forwards, multiple steps back-/forwards) and reliably working out if and what fragments swup should exchange (including skipping the animations if `animateHistoryBrowsing` is set to `true`), feels pretty impossible IMHO. 

Maybe a better route would be to finalize the API for leaving and re-entering the swup context, as proposed in #512 . This way, advanced consumers would be free to implement their own solutions for fragments/overlays etc... 

What do you think?

daun:
@hirasso I'd like to try one more thing — adding an API for defining which fragments to replace. Then we can see if it's intuitive or too bulky. I imagine that being some kind of callback function.

I can picture a combination of the `resolvePath` feature and the morph plugin with `onBeforeElementUpdated` callbacks  for building custom fragment support. Or a separate fragment plugin with a callback that would have some overlap with the morph plugin, but a slightly more focused/specialized API?

hirasso:
@daun I'm all ears! Looking forward to see what you come up with.

daun:
Closing in favor of the [fragment plugin](https://github.com/swup/fragment-plugin).

PR #504, opened 2022-09-03
Update readme

daun:
The readme recently felt a bit out of date. I've tried to clean it up a little, explain the motivation and main features and provide some useful links on the very top. The logo is a bit smaller here to allow us to explain that swup is very early on and avoid having to scroll down.

I've been thinking about recreating the example GIF in better resolution and with a browser UI around it, to make it a bit more appealing, but that's for later.

You can check it out by visiting the branch URL directly: [updated](https://github.com/swup/swup/tree/update-readme) vs. [original](https://github.com/swup/swup/tree/master) versions.

Screenshot 2022-09-03 at 17 16 54

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1,16 +1,20 @@
-<a href="https://github.com/swup/swup/discussions/424">We're looking for maintainers!</a>   👀
+<br>
 
-<br/>
+<p align="center">
+  <img width="280" alt="swup" src="https://swup.js.org/assets/images/swup-logo.svg">
+</p>
 
-<img src="https://swup.js.org/assets/images/swup-logo.svg" alt="swup logo" height="50" align="right">
+<p align="center">
+  <a href="https://www.npmjs.com/package/swup"><img src="https://img.shields.io/npm/v/swup.svg" alt="npm version"></a>
+  <img src="https://img.shields.io/bundlephobia/minzip/swup.svg" alt="Gzip Size">
+  <a href="https://github.com/gmrchk/swup/blob/master/LICENSE"><img src="https://img.shields.io/github/license/gmrchk/swup.svg" alt="License"></a>
+  <a href="https://www.npmjs.com/package/swup"><img src="https://img.shields.io/npm/dt/swup.svg" alt="npm downloads"></a>
+  <a href="https://circleci.com/gh/swup/swup"><img src="https://circleci.com/gh/swup/swup.svg?style=svg" alt="CircleCI"></a>
+</p>
+
+<br>
 
 # Swup
-
-[![npm version](https://img.shields.io/npm/v/swup.svg)](https://www.npmjs.com/package/swup)
-[![Bundle size](https://img.shields.io/bundlephobia/minzip/swup.svg)](https://bundlephobia.com/package/swup)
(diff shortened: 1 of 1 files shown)


hirasso:
💚 I like the simplification and cleanup of the text very much!

Regarding the header:

1. "We are looking for maintainers" at the very top reads to me a bit like "not maintained anymore" or "close to death". I would propose to move it to the very bottom, under the **Contributing** section. My guess is that is is more likely that we will find suitable future maintainers by contributions rather than by applications, anyways.
2. I would prefer the Swup logo to stay in the center and large. Also the slogan and the badges sit nicely there, IMHO. So basically the same as it is now, but without the maintainers message, the gray `<hr>` and, of course with "...static web**site**", I like this change a lot :) :

image

We could also think about adding Emojis to each of the bullet points, just because, you know, emojis are nice :) Something like...

- ✨ Auto-detects [CSS transitions](https://swup.js.org/getting-started/how-it-works) for perfect timing
- 🔗 Updates URLs and preserves native browser history behavior
- 📦 Uses a [cache](https://swup.js.org/api/cache) to speed up subsequent page loads
- 💡 Offers [events](https://swup.js.org/events) for hooking into the lifecycle
- 🔌 Has a powerful [plugin system](https://swup.js.org/plugins) and many official and third-party plugins
- 🎨 Provides ready-to-go [themes](https://swup.js.org/themes) to get started quickly

...what do you think? Too much?

daun:
Sure, I don't mind keeping the logo in focus and centered. Let's still reduce it in size a bit. Being surrounded by whitespace, I don't think it needs to be that large. On mobile, it's also a bit oversized.

I like emojis :) Maybe without the bullet points, then?

hirasso:
I implemented the discussed changes. I would prefer to keep the bullet points in the features list even if we have emojis, otherwise it wouldn't break nicely on mobile.

daun:
Great! I'll admit it looks much better with the logo centered. I've just restored the bundlephobia link and switched back to a nicer CircleCI batch. Other than that, happy to merge. Let's check with @gmrchk if he has any input.

hirasso (dismissed):

readme.md
```diff
@@ -1,44 +1,60 @@
...
-    - [Sites using swup](https://github.com/swup/swup/discussions/333)
-
-If you're having trouble implementing swup, checkout [Common Issues](https://swup.js.org/other/common-issues), [Closed Issues](https://github.com/gmrchk/swup/issues?q=is%3Aissue+is%3Aclosed) or open a [new one](https://github.com/gmrchk/swup/issues/new).
-
-[Become a backer or sponsor on Open Collective](https://opencollective.com/swup) or support swup through [GitHub sponsors](https://github.com/sponsors/gmrchk). 
+<br>
+
+# Swup
+
+Complete, flexible, extensible and easy to use page transition library for your static website.
```
gmrchk, line 19: Now that I read that part that I wrote myself, the "static" keyword is a bit misleading, since Swup supports dynamic web as well. Your description with server-side rendered is much better!

```suggestion
Complete, flexible, extensible, and easy-to-use page transition library for your server-side rendered website.
```

gmrchk (dismissed):

hirasso (approved):

PR #508, opened 2022-09-05
Add option `cache.resolvePath: path => path`

hirasso:
Allows control over how swup's cache resolves a path. Very simple usage example:

```js
const swup = const swup = new Swup({
    cache: {
        resolvePath: path => {
            if( path.startsWith('/projects/?') ) return '/projects/';
            return path;
        }
    }
});
```

### DIFF
--- a/src/modules/Cache.js
+++ b/src/modules/Cache.js
@@ -1,13 +1,25 @@
 import { getCurrentUrl, normalizeUrl } from '../helpers.js';
 
 export class Cache {
-	constructor() {
+	constructor(options) {
 		this.pages = {};
 		this.last = null;
+
+		const defaults = {
+			resolvePath: path => path
+		}
+		this.options = {
+			...defaults,
+			...options
+		};
+	}
+
+	normalizeUrl(url) {
+		return this.options.resolvePath(normalizeUrl(url));
 	}
 
 	cacheUrl(page) {
-		page.url = normalizeUrl(page.url);
+		page.url = this.normalizeUrl(page.url);
 		if (page.url in this.pages === false) {
 			this.pages[page.url] = page;
 		}
@@ -16,7 +28,7 @@ export class Cache {
 	}
 
 	getPage(url) {
-		url = normalizeUrl(url);
+		url = this.normalizeUrl(url);
 		return this.pages[url];
 	}
 
@@ -25,7 +37,7 @@ export class Cache {
 	}
 
 	exists(url) {
-		url = normalizeUrl(url);
+		url = this.normalizeUrl(url);
 		return url in this.pages;
 	}
 
--- a/src/modules/renderPage.js
+++ b/src/modules/renderPage.js
@@ -3,7 +3,7 @@ import { Link } from '../helpers.js';
 const renderPage = function(page, popstate) {
 	document.documentElement.classList.remove('is-leaving');
 
-	const isCurrentPage = this.getCurrentUrl() === page.url;
+	const isCurrentPage = this.cache.normalizeUrl(this.getCurrentUrl()) === page.url;
 	if (!isCurrentPage) return;
 
 	// replace state in case the url was redirected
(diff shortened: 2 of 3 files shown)


hirasso:
This is the first of a few new swup options I just discussed with @daun , to make it easier to work with custom handling of certain routes in a website. I'm struggeling to clearly explain it in writing, happy to walk you through our thought process in our next meeting @gmrchk

src/modules/Cache.js
```diff
@@ -1,13 +1,25 @@
...
 export class Cache {
-	constructor() {
+	constructor(options) {
 		this.pages = {};
 		this.last = null;
+
+		const defaults = {
+			resolvePath: path => path
+		}
+		this.options = {
```
daun, line 11: Does spreading cause an error if a user passes `true` or `false` for the `cache` option? I'd rather we add a check here for object types.

src/modules/Cache.js
```diff
@@ -1,13 +1,25 @@
...
+			...defaults,
+			...options
+		};
+	}
+
+	normalizeUrl(url) {
```
daun, line 17: Can we change the naming here? This step does two things: normale + resolve, it probably shouldn't be called just normalize. Or maybe the normalizeUrl helper itself should be renamed more clearly to indicate it's just stripping the hostname?

src/modules/renderPage.js
```diff
@@ -3,7 +3,7 @@ import { Link } from '../helpers.js';
 const renderPage = function(page, popstate) {
 	document.documentElement.classList.remove('is-leaving');
 
-	const isCurrentPage = this.getCurrentUrl() === page.url;
```
daun, line 6: I feel like this shouldn't be part of this PR. Resolving cache entries and resolving the current url is two different things, no?

src/modules/Cache.js
hirasso: both `{...true}` as well as `{...false}` yields an empty object, so we should be fine. The cache doesn't really need to know about if it is enabled or not. Still, maybe it would be clearer upon inspection to resolve the enabled state into the options. I will take care of that.

src/modules/Cache.js
hirasso: I agree, will change it to `resolveUrl`. `normalizeUrl` does also remove the hash if there is one. I would keep it as it is now, so that we don't get too many changes at once.

src/modules/renderPage.js
hirasso: This actually needs to be part of this PR, otherwise the new feature would not work.

daun: Mmh. Maybe I didn't understand the feature exactly? I thought this is about being able to cache similar results. I wouldn't expect a cache url resolver to resolve urls outside of the cache. If I have `/projects` and `/projects?filter=...` and both resolve to `/projects` in the cache, I would expect the cache to return the same html for either one; however I wouldn't expect it to stop rendering the new page entirely.

hirasso:
I just realized that `cache.resolvePath` must also be applied if detecting redirects, since the `responseURL` isn't necessary the URL we want to redirect to on popstate. I am getting the feeling that maybe `resolvePath` should be moved outside of the cache and into the swup main instance. Happy to discuss this, ideally together with @gmrchk , he should know all the implications best 💡

daun:
I guess my confusion then comes from this being proposed as a cache option when indeed it should be a global option...

hirasso:
Exactly. If you agree that this could be an improvement, I will try to refactor this to be a global option.

daun:
Yeah, let's rewrite this as a global option and see how it fits 🥯

hirasso:
What would you say, should I submit this feature to the `next` branch or to the current one? It will not be a breaking change but will it become too messy if we continue to add features to both branches? Need your git know-how here 🧙‍♀️

daun:
We could start submitting all new features to the next branch, and only submit bugfixes to the current master. That way, there won't be too much work in the 3.0 release.

hirasso:
Great, let's do this.

hirasso:
Closing in favor of #512

PR #511, opened 2022-09-10
Allow ignoring links from function callback

daun:
**Description**

Currently, the only official way of ignoring links is by appending the link selector option. This can get very complex very easily and it's not possible to do more advanced checks like e.g. ignoring all links inside a specific container. (The unofficial way would be listening to and preventing the click event in the capture phase).

The proposed change will split the check into to phases: select all links by selector, then filter via user-supplied callback.

This feature has been requested a few times and would be something I'd find very useful in my own projects.

**Advantages**

- Makes the selector much simpler
- Allows ignoring a whole container element with the `data-no-swup` attribute, not just every link individually
- We can add certain checks that will always skip swup, like `target=_blank` or the `download` attr on links
- Plugins can reuse some of the logic for other visit "triggers", e.g. forms plugin

**Checks**

- [x] The PR is submitted to the ~~`master`~~`next` branch
- [x] The documentation was [updated](https://github.com/swup/docs/pull/81) as required

**Additional information**

- This is a breaking change since we're ignoring `a[download]` elements. Hence submitting to the `next` branch.

**Questions**

- How would this tie in with #508 ?

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -27,9 +27,7 @@ export default class Swup {
 		let defaults = {
 			animateHistoryBrowsing: false,
 			animationSelector: '[class*="transition-"]',
-			linkSelector: `a[href^="${
-				window.location.origin
-			}"]:not([data-no-swup]), a[href^="/"]:not([data-no-swup]), a[href^="#"]:not([data-no-swup])`,
+			linkSelector: `a[href^="${window.location.origin}"], a[href^="/"], a[href^="#"]`,
 			ignoreLink: (el) => false,
 			cache: true,
 			containers: ['#swup'],
@@ -193,7 +191,17 @@ export default class Swup {
 		document.documentElement.classList.remove('swup-enabled');
 	}
 
+	isValidTrigger(triggerEl) {
+		if (triggerEl.closest('[data-no-swup]')) {
+			return false;
+		}
+		return true;
+	}
+
 	linkClickHandler(event) {
+		if (!this.isValidTrigger(event.delegateTarget)) {
+			return;
+		}
 		if (this.options.ignoreLink(event.delegateTarget)) {
 			return;
 		}
--- a/src/index.js
+++ b/src/index.js
@@ -30,6 +30,7 @@ export default class Swup {
 			linkSelector: `a[href^="${
 				window.location.origin
 			}"]:not([data-no-swup]), a[href^="/"]:not([data-no-swup]), a[href^="#"]:not([data-no-swup])`,
+			ignoreLink: (el) => false,
 			cache: true,
 			containers: ['#swup'],
 			requestHeaders: {
@@ -193,6 +194,10 @@ export default class Swup {
 	}
 
 	linkClickHandler(event) {
+		if (this.options.ignoreLink(event.delegateTarget)) {
+			return;
+		}
+
 		// no control key pressed
 		if (!event.metaKey && !event.ctrlKey && !event.shiftKey && !event.altKey) {
 			// index of pressed button needs to be checked because Firefox triggers click on all mouse buttons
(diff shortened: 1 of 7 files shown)


src/index.js
```diff
@@ -27,9 +27,7 @@ export default class Swup {
 		let defaults = {
 			animateHistoryBrowsing: false,
 			animationSelector: '[class*="transition-"]',
-			linkSelector: `a[href^="${
-				window.location.origin
-			}"]:not([data-no-swup]), a[href^="/"]:not([data-no-swup]), a[href^="#"]:not([data-no-swup])`,
+			linkSelector: `a[href^="${window.location.origin}"], a[href^="/"], a[href^="#"]`,
```
hirasso, line 30: This could be a good chance to also add the selector `a[href^="./"]`. That kind of link is necessary in combination with the [`<base>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base) element.

hirasso:
This looks great! 🥁

daun:
@hirasso Any reason you've removed the comment about adding support for relative links and the base tag? I think it's a nice goal to support everything the browser natively supports as well, so that adding swup can't break a site or force too many adjustments. For that to work, the selector would probably be just `a` and we'd check the protocol/hostname in JS.

hirasso:
Damn email notifications! ;) ...I removed the comment because I wasn't sure if it holds true that a link starting with `./something/` would always point towards the same context. I am using it in some of my test sites by extending the `selector` option and it works great, but I haven't really explored it to the end.

daun:
Huh. I didn't know `./something/` was a valid URL, I've always assumed `something/` to be the correct way of resolving from the current path. In any case, the nice thing is that the browser will just resolve all that to a complete URI with host and correct base. I think adding support for that would make sense.

daun:
Notes from today's call between @daun and @hirasso 

- Move `location.origin` check and `data-no-swup` check into default `ignoreLink` function
- Set `linkSelector` to `a`, deprecate it and stop mentioning it in the docs
- Instead, refer people to the `ignoreLink` option and document its default implementation
- New default behaviour going forward:
  - swup will detect clicks on all links, and filter them using the `ignoreLink` function
  - Legacy projects still using the `linkSelector` option will keep working as expected

daun:
Ready for review and some additional manual testing.

daun:
I've taken the liberty and reformatted some of the options.

daun:
Merged recent updates to the test suite, still passing.

hirasso:
Nevermind, it was my node version (`v18.12.1`) being too new. Passes with node `v16.13.2`.

daun:
🤠

src/index.js
```diff
@@ -251,6 +258,13 @@ export default class Swup {
 		}
 	}
 
+	triggerWillOpenNewWindow(triggerEl) {
```
hirasso, line 261: Why are you using `triggerWillOpenNewWindow` instead of `linkWillOpenNewWindow` here?

hirasso:
Has this been linted?

daun:
No 🤡 Will do it now.

src/index.js
daun: I'm trying to think ahead and make this usable for `form`s and the form plugin as well. Is this just making it confusing?

hirasso: No, that's what I was presuming. Just wanted to make sure. Looking good, will approve this now!

daun:
🚀

hirasso:
Very nice! At some point we should probably also think about cleaning up the rest of the `linkClickHandler` function. My brain always melts away if it has to parse more then one level of if...else 🫠

daun:
@hirasso It's actually already cleaned up a bit in the `fragments` branch.

Since we're not merging the fragments feature for now, I'll work on extracting some of the cleanup and restructuring and create a separate merge for those.

daun:
@hirasso I've created #535 to include some of the refactoring work from the fragments feature in the next branch.

Issue #236, opened 2020-01-03
Prevent navigation?


svennerberg:
In some cases I would like to prevent swup from navigating to the page that the clicked link refers to. Is that possible?

My idea was to  do something like this:

```javascript
swup.on('clickLink', e => {
  if (e.delegateTarget.attributes.disabled) {
    // prevent navigation
  }
})
```
The problem is that I can't find a way to stop swup from navigating. Any ideas?

gmrchk:
Hey,
this is not possible with swup in the form that you suggested. However, swup has an [API](https://swup.js.org/api/methods) that you can use to navigate between pages on your own, so you can implement any logic on top of it that you need.

```javascript
const link = ...;
link.addEventListener('click', event => {
    if (/* your logic here */) {
         swup.loadPage({
             url: event.target.href
         });
    }
});
```

FYI code not tested...

mskocik:
I agree with @svennerberg, it would be helpful, this would be possible to prevent swup navigation with 2 possible options:

- discard click event
- continue with default click event behaviour  

Imagine scenario, when custom pages can add additional handler to `clickLink` event, which could eventually cancel navigation to given page. 

My real use-case was is: I have html with _specific_ links (external, in specific context). These links don't and cannot have `data-no-swup` attribute, but I definitely need to prevent default swup navigation. This is my workaround where I additionally set `data-no-swup` attribute  and programmatically click on that link again.

```javascript
swup.on('clickLink', e => {
    const link = e.delegateTarget;
    const href = link.href;
    const fromTrix = link.closest('.trix-content');
    if (!fromTrix && (href.indexOf(window.location.host) > -1 || href.substr(-1) === '#')) return;
    link.href = window.location.pathname + window.location.search;
    link.dataset.swupSkip = true;   // setting data-no-swup attr to prevent swup navigation
    link.target="_blank";
    setTimeout(() => {
        link.href = href;
        link.click(); // link will be opened in new window now
    }, 20);
});
```

gmrchk:
Hi, sorry for the delay!

I get the point here, but it would a bit problematic and would introduce options only to specific swup event handlers which could be confusing, so let's try to solve the issue with the current setup.

Can we use a swup [link selector option](https://swup.js.org/options#link-selector) to select only links we want to use, eg. links in the header, footer, links in some other elements, but not the ones you want to be prevented?

mskocik:
[...]

PR #512, opened 2022-09-12
New global option: `resolvePath`

hirasso:
Replaces #508 

Adds a new global option to swup, that allows to run a callback for how swup should resolve a path. 

## What is the problem?

Right now, to be able to leave and re-enter control over the browser URL/state by swup, three steps are necessary: 

- Use [`skipPopStateHandling`](https://swup.js.org/options#skip-popstate-handling) with custom logic
- Manually ignore links with one of the given options ([`linkSelector`](https://swup.js.org/options#link-selector), [`stopPropagation()`](https://github.com/swup/swup/issues/236#issuecomment-1203786475) during the capture phase,...)
- Manually adjust the cache URLs so that back- and forward navigation doesn't need additional server requests

While this kind of behavior is already possible to achieve with the currently available swup API, it adds a lot of boilerplate to each new site that wants to have this. Also, the current cache implementation is less than ideal for that kind of scenario right now. It will either fill itself up with unnecessary duplicates of the same page or have cache misses for pages that were previously already loaded from the server.

## Proposed solution

All three of the above topics basically boil down to always the same question: Should various paths actually be resolved to just one by swup?

- on `popState`: Is the resolved version of the new browser URL the same as the previous one? > ignore
- on `clickLink`: Is the resolved href of the link the same as the current browser URL? > ignore
- Accordingly to swup ignoring the above events if paths resolve to the same one, the cache can also just keep one copy for all URLs that resolve to the same one.

## Demo & Test Site

I have set-up a test site here: https://test.rassohilber.com/swup-resolve-path/test/ 

On that site, the whole filter logic is done by a custom component that doesn't want swup to interfere. It does everything from updating the URL to rendering the state. Navigation away from the list to one of the character detail pages or to the "about"-page and back again is being handled by swup again. To make it more obvious that there is no page load between filters, I added a staggered animation to the items that only runs if the page is being fully rendered (initially and by swup).

## Code Example

```js
const swup = new Swup({
  resolvePath: path => {
    // resolve every URL that contains '/?=filter=xyz' to '/'
    return path.replace(/\/\?filter=.+/, '/');
  }
})
```

## Checks
- [x] The code was linted before pushing
- [x] All tests are passing
- [ ] Make sure that [restoring scroll positions in ScrollPlugin](https://github.com/swup/scroll-plugin/pull/26) also uses the resolved path

### DIFF
--- a/cypress/e2e/main.spec.js
+++ b/cypress/e2e/main.spec.js
@@ -2,7 +2,7 @@
 
 // window._swup holds the swup instance
 
-const durationTolerance = 0.1; // 10% plus/minus
+const durationTolerance = 0.15; // 10% plus/minus
 
 context('Window', () => {
     beforeEach(() => {
--- a/cypress/e2e/main.spec.js
+++ b/cypress/e2e/main.spec.js
@@ -276,4 +276,44 @@ context('Window', () => {
             });
         });
     });
+
+	it('should ignore links for equal resolved paths', () => {
+		 cy.window().then(window => {
+			window._swup.options.resolvePath = path => '/page1/';
+			cy.triggerClickOnLink('/page2/');
+			cy.shouldBeAtPage('/page1/');
+		 });
+	});
+
+	it('should skip PopState handling for equal resolved paths', () => {
+		cy.window().then(window => {
+			window._swup.options.resolvePath = path => '/page1/';
+			window.history.pushState(
+				{
+					url: '/pushed-page-1/',
+					random: Math.random(),
+					source: 'swup'
+				},
+				document.title,
+				'/pushed-page-1/'
+			);
+
+            window.history.pushState(
+                {
+                    url: '/pushed-page-2/',
+                    random: Math.random(),
+                    source: 'swup'
+                },
+                document.title,
+                '/pushed-page-2/'
+            );
+
+			cy.wait(500).then(() => {
+				window.history.back();
+				cy.shouldBeAtPage('/pushed-page-1/');
+				cy.shouldHaveH1('Page 1');
+			})
+		});
+	})
+
 });
--- a/src/index.js
+++ b/src/index.js
@@ -274,7 +274,7 @@ export default class Swup {
 	/**
 	 * Utility function to validate and run the global option 'resolvePath'
 	 * @param {string} path
-	 * @returns string the resolved path
+	 * @returns {string} the resolved path
 	 */
 	resolvePath(path) {
 		if( typeof this.options.resolvePath !== 'function' ) {
@@ -283,7 +283,7 @@ export default class Swup {
 		return this.options.resolvePath(path);
 	}
 	/**
-	 * Compares the resolved path of two paths and returns true if they are the same
+	 * Compares the resolved version of two paths and returns true if they are the same
 	 * @param {string} path1
 	 * @param {string} path2
 	 * @returns {boolean}
--- a/src/index.js
+++ b/src/index.js
@@ -277,9 +277,15 @@ export default class Swup {
 	 */
 	resolvePath(path) {
 		if( typeof this.options.resolvePath !== 'function' ) {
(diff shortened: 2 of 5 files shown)


daun:
Nice! Will check it out this week.

daun:
One thought beforehand: returning `null` from the resolver function will result in a warning. I think returning `null` or `undefined` could have a meaning here as well, i.e. "leave the url as-is". Probably no need to throw a warning, then?

hirasso:
I would actually prefer to be quite strict here, a warning will hopefully help people with the implementation. The function should *always* return a non-empty string. I actually also thought about validating if the resolved path is actually a path and not a full URL, as well.

hirasso:
Could you also have a look at the tests? They are passing locally, I don't understand what's going on in the CI 🙏

daun:
Looks like we reached the monthly limit for the free CircleCI plan, it won't show the test run :(

hirasso:
I think it might be the complex transitions test, the duration was a bit outside the expected range one time I let it run locally.

daun:
You could try increasing the expected range — it's a constant somewhere at the top of `main.spec.js`, I think `0.1` at the moment. Setting it to 0.15 or so might already help?

hirasso:
Great, that worked! 🎺

hirasso:
@daun do you have any reservations against merging this?

daun:
@hirasso Only that I've been super busy and haven't had the time to thoroughly test this. I hope I can give it a spin later this week or on the weekend.

daun:
Sorry for the delay in checking this out. Looks good! It's something we should definitely have in the codebase.

One question about the scope: does the linked example always return the same HTML, regardless of the `filter` param? I'm thinking about the implications of resolving the cache urls as well here.

hirasso:
> One question about the scope: does the linked example always return the same HTML, regardless of the filter param? I'm thinking about the implications of resolving the cache urls as well here.

Yes, each `?filter=...`-page in the example returns the same dataset with all items, regardless of the filter in the URL. The pages only differ in two things that are good SEO practice – the initial document title reflects the selected filter and the items that match the filter are being rendered initially (and then thrown away as soon as my component takes over).

The cache *must* use `resolveURL`, as well, if you think about the user journey: They have changed the filter a few times (swup ignores these), then navigated to the "about" page (wich is handled by swup again). If they now would go back in the history, they would expect the previous pages to be in the cache. Without the cache also using `resolveURL`, there would be no cache records and swup would needlessly have to re-fetch the page from the server.

If you need further clarification, I would be happy to walk you through it in a call – otherwise I would proceed and merge this. Documentation can be a separate PR.

hirasso:
Actually, thinking about it: The concept is actually quite similar to [canonical URLs](https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls):

> A canonical URL is the URL of the page that Google thinks is most representative from a set of duplicate pages on your site. 

So maybe renaming the option to `getCanonicalURL` would help with understanding it better, what do you think?

daun:
Let's jump on a quick call whenever convenient! I agree with the reasoning but would just like to get your input on how I have been solving those filter ui situations. Ideally this PR would cover both ways of tackling it.

I'd stick with the current baming. Calling it _canonical_ would somewhat imply being tied to a hypothetically present meta tag whereas _resolve_ is more neutral.

hirasso:
I see... Do you also prefer `resolvePath` over `mapPath`?

daun:
I think I prefer `resolve` over `map` as well. To me, mapping implies a 1:1 relationship whereas resolving implies combining as well as rewriting. Neither sounds false, but `resolve` is the one that'd make me instantly understand the concept when encountering the option in the docs.

How strong is your aversion to using `resolve` here?

hirasso:
Not strong it all. Just wanted to make sure we are naming it correctly. Let's stay with resolve, then! 👏

PR #535, opened 2022-12-07
Add refactorings from fragments branch

daun:
The fragments feature won't be part of swup 3, but the feature branch has a few things we'd like to merge already:

- General refactoring, simplifications and formatting
- Add early returns to event handlers to avoid excessive nesting of if/else blocks
- Extract `leavePage` and `enterPage` into separate modules/steps, making work on future features easier
- Add `nextTick` helper for animation timings
- Add `updateHistoryRecord` helper in addition to existing `createHistoryRecord`

Tests are still passing.

### DIFF
--- a/src/helpers/createHistoryRecord.js
+++ b/src/helpers/createHistoryRecord.js
@@ -1,13 +1,12 @@
-const createHistoryRecord = (url) => {
-	window.history.pushState(
-		{
-			url: url || window.location.href.split(window.location.hostname)[1],
-			random: Math.random(),
-			source: 'swup'
-		},
-		document.title,
-		url || window.location.href.split(window.location.hostname)[1]
-	);
+const createHistoryRecord = (url, customData = {}) => {
+	url = url || window.location.href.split(window.location.hostname)[1];
+	const data = {
+		url,
+		random: Math.random(),
+		source: 'swup',
+		...customData
+	};
+	window.history.pushState(data, document.title, url);
 };
 
 export default createHistoryRecord;
--- a/src/helpers/createHistoryRecord.js
+++ b/src/helpers/createHistoryRecord.js
@@ -1,5 +1,7 @@
+import { getCurrentUrl } from '../helpers.js';
+
 const createHistoryRecord = (url, customData = {}) => {
-	url = url || location.href;
+	url = url || getCurrentUrl({ hash: true });
 	const data = {
 		url,
 		random: Math.random(),
--- a/src/helpers/createHistoryRecord.js
+++ b/src/helpers/createHistoryRecord.js
@@ -1,12 +1,12 @@
 const createHistoryRecord = (url, customData = {}) => {
-	url = url || window.location.href.split(window.location.hostname)[1];
+	url = url || location.href;
 	const data = {
 		url,
 		random: Math.random(),
 		source: 'swup',
 		...customData
 	};
-	window.history.pushState(data, document.title, url);
+	history.pushState(data, '', url);
 };
 
 export default createHistoryRecord;
--- a/src/helpers/updateHistoryRecord.js
+++ b/src/helpers/updateHistoryRecord.js
@@ -1,13 +1,13 @@
 const updateHistoryRecord = (url = null, customData = {}) => {
-	url = url || window.location.href; // window.location.href.split(window.location.hostname)[1];
+	url = url || location.href;
 	const data = {
-		...window.history.state,
+		...history.state,
 		url,
 		random: Math.random(),
 		source: 'swup',
 		...customData
 	};
-	window.history.replaceState(data, document.title, url);
+	history.replaceState(data, '', url);
 };
 
 export default updateHistoryRecord;
--- a/src/helpers/updateHistoryRecord.js
+++ b/src/helpers/updateHistoryRecord.js
@@ -1,0 +1,13 @@
+const updateHistoryRecord = (url = null, customData = {}) => {
+	url = url || window.location.href; // window.location.href.split(window.location.hostname)[1];
+	const data = {
+		...window.history.state,
+		url,
+		random: Math.random(),
+		source: 'swup',
+		...customData
+	};
+	window.history.replaceState(data, document.title, url);
+};
+
+export default updateHistoryRecord;
--- a/src/helpers/updateHistoryRecord.js
+++ b/src/helpers/updateHistoryRecord.js
@@ -1,5 +1,7 @@
+import { getCurrentUrl } from '../helpers.js';
+
 const updateHistoryRecord = (url = null, customData = {}) => {
-	url = url || location.href;
+	url = url || getCurrentUrl({ hash: true });
 	const data = {
 		...history.state,
 		url,
--- a/src/index.js
+++ b/src/index.js
@@ -32,10 +32,9 @@ export default class Swup {
 			animationSelector: '[class*="transition-"]',
 			cache: true,
 			containers: ['#swup'],
-			ignoreLink: (el) => (
-				el.origin !== window.location.origin ||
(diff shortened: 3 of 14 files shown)


src/helpers/createHistoryRecord.js
```diff
@@ -1,13 +1,12 @@
...
-		{
-			url: url || window.location.href.split(window.location.hostname)[1],
-			random: Math.random(),
-			source: 'swup'
-		},
-		document.title,
-		url || window.location.href.split(window.location.hostname)[1]
-	);
+const createHistoryRecord = (url, customData = {}) => {
+	url = url || window.location.href.split(window.location.hostname)[1];
```
hirasso, line 2: While we are at it, couldn't we use swups `getCurrentUrl()` here?
```js
import getCurrentUrl from "./getCurrentUrl";
const createHistoryRecord = (url, customData = {}) => {
	url = url || getCurrentUrl();
    // ...
}
```

src/helpers/updateHistoryRecord.js
```diff
@@ -0,0 +1,13 @@
+const updateHistoryRecord = (url = null, customData = {}) => {
+	url = url || window.location.href; // window.location.href.split(window.location.hostname)[1];
```
hirasso, line 2: Same as with `createHistoryRecord`. Or what was the reason the full URL is being used here?

src/helpers/createHistoryRecord.js
hirasso: Also, I always found `getCurrentUrl` always a bit confusing, naming-wise. How about we change it to `getCurrentPath` and add an alias for backwards-compatibility, what do you think?

src/index.js
```diff
@@ -196,65 +201,57 @@ export default class Swup {
...
+				if (element) {
+					updateHistoryRecord(url + hash);
+				} else {
+					console.warn(`Element for offset not found (#${hash})`);
 				}
 			}
-		} else {
-			// open in new tab (do nothing)
-			this.triggerEvent('openPageInNewTab', event);
+		} else if (!this.isSameResolvedPath(url, getCurrentUrl())) {
```
hirasso, line 248: Here would be another chance to do an early returns. If you agree, I could push the change directly to this branch.

src/modules/loadPage.js
```diff
@@ -1,100 +1,66 @@
...
 
 	this.triggerEvent('transitionStart', popstate);
 
 	// set transition object
-	if (data.customTransition != null) {
-		this.updateTransition(window.location.pathname, data.url, data.customTransition);
-		document.documentElement.classList.add(`to-${classify(data.customTransition)}`);
-	} else {
-		this.updateTransition(window.location.pathname, data.url);
+	this.updateTransition(window.location.pathname, url, customTransition);
```
hirasso, line 13: I know, not related to this PR, but shouldn't we use `getCurrentUrl` here as well?
```js
this.updateTransition(getCurrentUrl(), url, customTransition);
```

hirasso (commented):
Very nice!! 🍰 Added a few comments/questions.

src/helpers/createHistoryRecord.js
daun: Done :) I'd prefer not to touch the naming, just maintenance-wise.

src/helpers/updateHistoryRecord.js
daun: Done :) It was just inconsistent since it doesn't matter in this context if the URL is relative or absolute.

src/index.js
daun: This one I'm unsure about. Do we exit early here before the first `if` check or after? Maybe you tackle that one.

src/modules/loadPage.js
daun: Done :) It has the added benefit that now the query params end up in the transition object as well.

src/index.js
hirasso: Ok, I'll take care of this.

hirasso: Did some more refactoring. Looking forward to hear what you think! Tests are passing. I also ran `lint` on the code, which turned the option `ignoreLink` to this:

```js
let defaults = {
	//...
	ignoreLink: (el) => 
		el.origin !== window.location.origin || el.closest('[data-no-swup]'),
	//...
};
```
It becomes slightly more readable by an explicit return IMHO:
```js
let defaults = {
	// ...
	ignoreLink: (el) => {
		return el.origin !== window.location.origin || el.closest('[data-no-swup]');
	},
	// ...
};
```

src/helpers/createHistoryRecord.js
hirasso: Hmm. I really find it confusing that we are using `getCurrentUrl()`, but then we also use `resolvePath` on it. We should probably either 

1. rename `resolvePath()` to `resolveUrl()`
2. or rename `getCurrentUrl()` to  `getCurrentPath()`

Since the thing actually _is_ a relative URL composed by `pathname` + `searchParams`, and also for backwards-compatibility, I would tend to rename the new option #1 then. What do you think?

src/index.js
daun: The reason I left it unlinted before is for readability. As it is, the one-liner works fine, but once you add more checks, the previous one-check-per-line-format is a lot more readable in my opinion, especially when explaining it in the docs. How do you feel about reverting it and adding a `prettier-ignore` above? Too fussy?

src/helpers/createHistoryRecord.js
daun: In that case, I also prefer option 1, renaming the new option `resolveUrl`. Could you do that for the main repo and the docs? And the scroll plugin, I guess, since that's already using it.

src/index.js
hirasso: > The reason I left it unlinted before is for readability. As it is, the one-liner works fine, but once you add more checks, the previous one-check-per-line-format is a lot more readable in my opinion, especially when explaining it in the docs. How do you feel about reverting it and adding a `prettier-ignore` above? Too fussy?

Hmm, not a fan of adding prettier-ignore there. I would then maybe prefer walking each condition with guard clauses. This will make it VERY clear what's happening. We could even add inline documentation:
```js
let defaults = {
	// ...
	ignoreLink: (el) => {
		// ignore external links:
		if (el.origin !== window.location.origin) return true;
		// ignore links that are inside a container with the attribute [data-no-swup]:
		if (el.closest('[data-no-swup]')) return true;
		// handle all other links
		return false;
	},
	// ...
};
```

daun:
@hirasso Ready to merge? 🧃

src/index.js
hirasso: Yes we shall! :)

hirasso: ...ready to merge then!

daun: 🚀

src/helpers/updateHistoryRecord.js
```diff
@@ -0,0 +1,15 @@
+import { getCurrentUrl } from '../helpers.js';
+
+const updateHistoryRecord = (url = null, customData = {}) => {
+	url = url || getCurrentUrl({ hash: true });
```
hirasso, line 4: Why did you add the optional hash here? Missed that before

hirasso: Sorry, wrong context... why does `getCurrentUrl()` now accept returning the hash as well?

daun: `getCurrentUrl` by default returns the url without a hash, but updating the history record should keep any current hash. I figured it's clearner to make it an option in `getCurrentUrl` than to combine `getCurrentUrl() + location.hash` or something. Should we rethink that? Not too late 🤡

PR #548, opened 2022-12-14
Create and export helper for event delegation

daun:
- Create a helper for delegating events
- Re-export as helper to allow plugins to delegate events without caring about event delegation libraries or versions
- Return our own abort controller to offer a stable API for removing listeners (every major version of `delegate-it` in recent months has changed the API for aborting event listeners)

```js
import { delegateEvent } from 'swup/helpers';

const delegation = delegateEvent('form',  'click', () => this.submitForm());

delegation.destroy();
```

**Question**

- Should we add this as a helper to the swup instance?
- Plugins could then just do `this.swup.delegateEvent(...)` without importing the helper at all
- Advantage: UMD browser builds will become smaller since they won't have to include `delegate-it`

**Future notes**

If/when we decide to update `delegate-it` to v5, the below would be the correct way of removing a listener. Holding off for now since v5 decreases browser support without adding features or fixes.

```js
const controller = new AbortController();
delegate(base, selector, type, callback, { ...eventOptions, signal: controller.signal });
return { destroy: () => controller.abort() };
```


daun:
@hirasso What do you think about this? How do you feel about including the helper for plugins to use without bundling their own version of delegate-it?

hirasso:
@daun sorry for the wait, I'm too busy to test it thoroughly on my end right now. Either you wait 2 more days (hopefully) or I'll just approve the PR. I have no objections :)

daun:
@hirasso There's no rush, just thought it might have fallen through the cracks. If you have no objections in principle, I'd like to add the helper on the swup instance and let you approve it then.

daun:
@hirasso I could also just merge it myself; it's going into `next` anyway.

hirasso:
Great, do that! :)

Issue #551, opened 2022-12-18
Handle non-http protocols with new `ignoreLinks` option


daun:
The upcoming implementation of `ignoreLinks` has one flaw: it doesn't ignore links with protocols like `tel:` or `mailto:`. We'll need to add that check before the release and add tests as well.

Edit: Actually, it does. We're checking for `link.origin` and comparing to `location.origin`, that includes the protocol. Still, we should make sure it's either http or https in any case.

```js
new URL(document.URL, document.baseURI)
new URL(window.location.pathname, document.baseURI)
```

Some questions to ask:

- Should we move the `origin` check into a separate function that cannot be overriden? I'm actually not sure what the advantage is of allowing cross-origin requests
- Should we switch to the native `URL` object instead of swup's `Link` class? Makes more sense to me
- Maybe look for inspiration in [Turbo's URL module](https://github.com/hotwired/turbo/blob/main/src/core/url.ts)

hirasso:
Right! Good catch!

hirasso:
Finally found some time investigating this. Indeed, `el.origin` is `null` for `mailto:` or `tel:` links.

> Should we move the origin check into a separate function that cannot be overriden? I'm actually not sure what the advantage is of allowing cross-origin requests

Yes, let's do that. We could just do the `origin` check directly before the `this.triggerWillOpenNewWindow(linkEl)` conditional and return early if it doesn't match

> Should we switch to the native URL object instead of swup's Link class? Makes more sense to me

It really feels like swups `Link` class only exists because of legacy browsers. If we are fine with [the browser support](https://caniuse.com/url), I don't have any other objections against using the `URL`API.

PR #554, opened 2022-12-21
Make `ignoreLink` accessible for plugins

hirasso:
- move all checks out of `linkClickHandler` and in a dedicated method `ignoreLink`
- move the `el.origin` check from the option inside this method

Makes it possible for plugins to call `swup.ignoreLink(el)`

Related: #551

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -211,9 +211,10 @@ export default class Swup {
 
 	linkClickHandler(event) {
 		const linkEl = event.delegateTarget;
+		const link = new Link(linkEl);
 
 		// Exit early if the link should be ignored
-		if (this.shouldIgnoreVisit(linkEl.href, { el: linkEl })) {
+		if (this.shouldIgnoreVisit(link.getHref(), { el: linkEl })) {
 			return;
 		}
 
@@ -231,7 +232,6 @@ export default class Swup {
 		this.triggerEvent('clickLink', event);
 		event.preventDefault();
 
-		const link = new Link(linkEl);
 		const url = link.getAddress();
 		const hash = link.getHash();
 
--- a/src/index.js
+++ b/src/index.js
@@ -195,7 +195,7 @@ export default class Swup {
 			return true;
 		}
 
-		// Ignore if the link/form would open new window (or none at all)
+		// Ignore if the link/form would open a new window (or none at all)
 		if (el && this.triggerWillOpenNewWindow(el)) {
 			return true;
 		}
@@ -205,7 +205,7 @@ export default class Swup {
 			return true;
 		}
 
-		// Finally, allow the link
+		// Finally, allow the visit
 		return false;
 	}
 
--- a/src/index.js
+++ b/src/index.js
@@ -30,9 +30,7 @@ export default class Swup {
 			animationSelector: '[class*="transition-"]',
 			cache: true,
 			containers: ['#swup'],
-			ignoreLink: (el) => {
-				return el.closest('[data-no-swup]');
-			},
+			ignoreVisit: (href, { el }) => el?.closest('[data-no-swup]'),
 			linkSelector: 'a[href]',
 			plugins: [],
 			resolveUrl: (url) => url,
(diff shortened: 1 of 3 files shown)


hirasso:
> 1 — Let's rename it to sound like a conditional: shouldIgnoreLink maybe. Right now it sounds like calling it will make swup ignore a specific link.

Good call, let's prefix the function with `should`!

Regarding your suggestion: You definitely have a broader oversight over swups plugin ecosystem. I'm not able to fully think it through without seeing the actual implementation – but your reasoning sounds plausible, would you like to give it a shot?

```js 
shouldIgnoreVisit({ href, el }) {
    // ...
}
```
How would you imagine that this would be called? Intuitively would have gone for a required `href` and an optional `el` parameter, something like this:
```js 
shouldIgnoreVisit(href, el = null) {
    // ...
}
```

daun:
Sure, I'll give it a try.

The reasoning behind the destructured object argument was to be able to add parameters later if required. But I like the required href param, so let's do both. That way we can call it with just a string as well.

```js
shouldIgnoreVisit(href, { el } = {}) {

}
```

hirasso:
Nice! Looking forward.

daun:
I've pushed a few changes.

- Rename `shouldIgnoreVisit` method
- Rename `options.ignoreLink` option to `ignoreVisit` to make its purpose clearer. Need to change docs as well, and your PR on the preload plugin.
- Pass around the `el` as optional parameter

Now we can check for forms as well:

```js
const form = event.delegateTarget;
const href = form.action;
if (this.swup.shouldIgnoreVisit(href, { el: form }) {
  //
}
```

I've used a `URL` object here for the origin check, I'd like to do a refactor to use that exclusively but that's another PR :)

daun:
What's great about this PR is any plugin can now extend and overwrite this method on the swup instance.

hirasso:
Looking good! I'll test it tomorrow and report back.

src/index.js
```diff
@@ -30,9 +30,7 @@ export default class Swup {
 			animationSelector: '[class*="transition-"]',
 			cache: true,
 			containers: ['#swup'],
-			ignoreLink: (el) => {
-				return el.closest('[data-no-swup]');
-			},
+			ignoreVisit: (href, { el }) => el?.closest('[data-no-swup]'),
```
hirasso, line 33: Shouldn't we make the second (options) parameter optional here as well?

```js
ignoreVisit: (href, { el } = {}) => el?.closest('[data-no-swup]'),
```

daun: Sure, why not. Feel free to push :)

hirasso:
Good to go from my side :) I'll move over to preload plugin then to adjust it the new method name.

daun:
Great! Can you merge? I'm away from the keyboard until the end of next week.

hirasso:
These duration tests are always making trouble in the ci... 🤖🐌

hirasso:
> Great! Can you merge? I'm away from the keyboard until the end of next week.

I can't merge without your approval. But no rush, the tests are failing again anyways :)

daun (approved):

hirasso:
So it seems to me like the tests sometimes  fail because the whole duration is *sometimes* the duration plus the delay and sometimes only the duration. Could this be the cause somehow? I find it quite hard to debug the tests...

hirasso:
Did some more investigation. The implementation of the transition animation seems off to me. I created a manual test to understand it's current behavior: It starts the animation with a duration of 400ms after a delay of 300ms: 
```css
animation: test-animation 400ms 300ms;
```
The animation fades the content out and in again:
```css
@keyframes test-animation {
  0%, 100% {
    transform: translateY(0);
    opacity: 1;
  }
  50% {
    transform: translateY(10px);
    opacity: 0;
  }
}
```
This results in the container being faded out and immediately in again, and only after the animation finished it replaces the content in the container. Not what we want, right? :)

I would presume the way we want keyframe animations to be used with swup would rather look something like this:
```css
@keyframes test-animation {
  from {
    transform: translateY(0);
    opacity: 1;
  }
  to {
    transform: translateY(10px);
    opacity: 0;
  }
}

html.is-changing .transition-main {
  animation: test-animation 300ms;
}
html.is-rendering .transition-main {
  animation-direction: reverse;
}
```
For the test case we could add the `animation-delay` back to the animation, like this:
```css
html.is-changing .transition-main {
  animation: test-animation 300ms 200ms; /* should result in 500ms total duration */
}
```

hirasso:
Tests are passing now, I'll merge and we can tackle the tests in another PR should we need to.

daun:
👽

PR #555, opened 2022-12-22
Switch to native URL API

daun:
**Examples**

```js
const { url, hash } = Location.fromElement(linkEl);
const { origin } = Location.fromUrl(history.state.url);
```

**Details**

- Use browser's native `URL` class for parsing URL parts
- Create a new `Location` class that returns an augmented `URL` instance
  - The reason for renaming it instead of keeping the class `Link` is to preferably have any plugins that use it fail at build time due to a missing import, rather than have it fail at runtime because of a missing method
- Add static methods for instantiating from either a URL string or DOM element
- Define a new `url` getter for the local path + query params
  - the previously used `getAddress()` was a bit hard to understand as address is not a common term in URLs
- Using getters simplifies the code quite a bit and allows for nice destructuring, see example above

**Implementation notes**

- Implementation as  ES6 class (`class Location extends URL {}`) will increase the bundle size by 10% (0.5K) and add tons of Babel class boilerplate for classes and getters
- For now we're just augmenting the `URL` object to keep bundle size low
- In the future, we can consider moving to inheritance as originally planned

Related to

#551


daun:
@hirasso How do you feel about the address getter in the url objects? It feels a bit nasty but I think we need some kind of abstraction. Maybe call it 'localPath' or 'localHref' or something more obvious. Or ceate another helper?

hirasso:
I haven't yet fully understood the problems this PR is trying to solve, and neither the implications it will have for backwards compatibility. Will need some time to review it after the holidays, or we meet live and you talk me through.

daun:
@hirasso No rush, let's talk whenever convenient after the holidays.

src/helpers/createUrl.js
```diff
@@ -0,0 +1,21 @@
...
+	if (hrefOrElement instanceof Element) {
+		href = hrefOrElement.getAttribute('href') || hrefOrElement.getAttribute('xlink:href');
+	} else {
+		href = hrefOrElement.toString();
+	}
+
+	const url = new URL(href, document.baseURI);
+
+	// Define custom 'address' getter for pathname + query params
+	Object.defineProperty(url, 'address', {
```
hirasso, line 12: I don't have any objections against extending the URL object with an `address` getter. This would also work, and I think is a little easier to parse:
```js
class ExtendedURL extends URL {
	get address() {
		return this.pathname + this.search;
	}
}
return new ExtendedURL(href, document.baseURI);
```

src/index.js
```diff
@@ -17,6 +17,7 @@ import {
 	cleanupAnimationClasses,
 	delegateEvent,
 	getCurrentUrl,
+	createUrl,
 	Link,
```
hirasso, line 21: This needs to be removed, the `Link` helper won't be there anymore, right? :)

hirasso (commented):
I like your proposed unification of handling URLs very much! Added two inline comments.

src/helpers/createUrl.js
hirasso: We could also call it `SwupURL` 🧋

hirasso:
A thing to discuss: Aren't we introducing breaking changes here? If so, is it worth it?

hirasso:
The `address` naming always felt a bit unfortunate (un-self-explanatory) to me.  Maybe we could use this chance to rename it to something that makes sense, if we decide to introduce this breaking change:

- `pathnameAndSearchParams`?
- `pathAndQuery`?
- `swupUrl`?
- a simple `url`?

The whole `createUrl` component could look like this then:

```js
class ExtendedURLObject extends URL {
	get url() {
		return this.pathname + this.search;
	}
}

function createUrl(hrefOrElement) {
	let href;
	if (hrefOrElement instanceof Element) {
		href = hrefOrElement.getAttribute('href') || hrefOrElement.getAttribute('xlink:href');
	} else {
		href = hrefOrElement.toString();
	}

	return new ExtendedURLObject(href, document.baseURI);
}

export default createUrl;
```

That way, code like this could become quite simpler to understand, because now `createUrl().url` would actually fit the naming of `getCurrentUrl()` or `resolveUrl()`:
```js
const url = createUrl(linkEl).url;
const hash = createUrl(linkEl).hash;

// Handle links to the same page and exit early, where applicable
if (!url || url === getCurrentUrl()) {
	this.handleLinkToSamePage(url, hash, event);
	return;
}
```

hirasso:
Maybe we could call the function `parseUrl` instead of `createUrl`?

```js
const { url, hash } = parseUrl(el.href);
```

Sorry long train ride, I've got too much time ;)

daun:
I really like the last solution. Calling the accessor `relativeUrl` or `localUrl` would be the most expressive name. But just `url` is gine since we're using that instead of `path` in most other places in the codebase as well.

daun:
I'm a bit torn between `parseUrl`, `createUrl` and a static `from` method on the URL object itself. In that case, `Location` could make sense as a name of the class.

```js
class Location extends URL {
  static from(hrefOrElement) {
    /* create instance */
    return new Location(href, document.baseURI);
  }
}

const { url, hash } = Location.from(href);

// or Location.create() or Location.parse()
```

hirasso:
How would the full component look like then?

hirasso:
I gave it a shot:

```js
/**
 * A helper for creating a Location from either an element
 * or a URL object/string
 */
export default class Location extends URL {
	/**
	 * Instantiate a Location from an element
	 * @param {Element} el
	 */
	static fromElement(el) {
		return new Location(href, document.baseURI);
	}
	/**
	 * Instantiate a Location from a URL object or string
	 * @param {URL|string} url
	 */
	static fromUrl(url) {
		return new Location(url.toString(), document.baseURI);
	}
	/**
	 * Get the URL that is being used throughout swup
	 * for identifying pages: pathname + searchParams
	 * @return string
	 */
	get url() {
		return this.pathname + this.search;
	}
}
```
As a starting point for further discussion I introduced two different `from*` factory functions, one for an element and one for a URL object or string. I would like this explicit approach better then the "on catch all" function that was being used before, since it makes everything clearer and would make an eventual port of swup to typescript simpler. What do you think?

This is fun! 👾

hirasso:
All this code designing really is fun. We'll still need to discuss how we get all plugins to use this, though. [Preload Plugin](https://github.com/swup/preload-plugin), for example, uses the old `Link` helper.

daun:
Awesome 🤠 I'm fine with keeping it explicit in the codebase but I'd prefer having a generic `from` method with an if/else block as well, just to clarify intent and simplify migrating to the new version. Typescript accepts union types for parameters, so that's still possible here I think.

Regarding plugins: we could either A) remove the `Link` class in swup 3 and port the plugins as well or B) deprecate it but keep it around to remove in the future. Both are fine, although I orefer A) to keep things simple. We can add a note to the v3 migration docs.

Or: option C) rename the new `Location` class `Link`, keep it around but switch to getters instead of methods.

What do you think?

hirasso:
I think we'll need to talk it through :)

hirasso:
Soooo... IF we actually decide to accept braking changes for swup@3, I'd like to propose that we rename every instance of `pathname` + `search` with `route`! I would then even want  to rename `getCurrentUrl` and `resolveUrl` to `getCurrentRoute` and `resolveRoute` 💥🥳

I notoriously have a hard time understanding intuitively what `url`/`address` etc. actually means in swup, I always have to look at the actual implementation. `route` would be a well-established and easy to reason-about term IMHO.

What do you think? :)

daun:
I like route as a name! Let's discuss the implications of that on a call soon.

daun:
TLDR; Ready for review 🤠

@hirasso I've had to go back to a custom getter on the `URL` object to avoid Babel adding 10% to the bundle size. I've left a comment in the codebase explaining that and a suggestion for a future inheritance-based implementation.

Anyway, I've taken your earlier comment about implementation speed to heart and I'd suggest just going with it. I've spent way too much time on this already 🤡 The breaking change is worth it in my opinion, for the use of destructuring in the codebase alone. The old `Link` class is only used internally or by plugins and not by sites, so I wouldn't worry here. Happy to port the plugins myself as well.

Regarding the `route` naming: Upon a little more thinking, we should discuss this separately from this PR since it involves multiple places. Personally, I don't think it's worth the trouble if we make sure to stick with `url` as local URL and `href` as absolute URL and get rid of `address`.

src/helpers/index.js
```diff
@@ -5,7 +5,6 @@ export { default as delegateEvent } from './delegateEvent.js';
 export { default as getDataFromHtml } from './getDataFromHtml.js';
 export { default as fetch } from './fetch.js';
 export { default as getCurrentUrl } from './getCurrentUrl.js';
-export { default as normalizeUrl } from './normalizeUrl.js';
+export { default as Location } from './Location.js';
```
hirasso, line 8: Something I wanted to ask before already: Why are we not simply importing the defaults like this?

```js
export Location from './Location.js';
```

hirasso (approved):
Tests are passing. Let's goooo 🎺

src/helpers/index.js
daun: I don't think that's valid syntax. Just checked and it errors out: `Declaration or statement expected`

daun: Turns out your example is [a proposed syntax](https://github.com/tc39/proposal-export-default-from) and there's [a babel plugin](https://babeljs.io/docs/en/babel-plugin-proposal-export-default-from) for it.

hirasso: Ooops! As you might have noticed in my comment ("[...]*importing* the defaults[...]"), I totally overlooked that these are `export`, not `import` statements. I saw an import using the `{ default as x }` syntax [over here](https://github.com/swup/swup/blob/next/src/helpers/createHistoryRecord.js#L1) the other day. That's where I was confused. But still good to know that my example is a proposed syntax! 😄

Issue #540, opened 2022-12-09
Optimize CI setup


daun:
- Avoid CircleCI usage limits (test runs are hidden when free quota is met)
- Speed up test runs (GitHub actions looks to be 50-70% faster than CircleCI)
- Being worked on in
  - #559 
  - #564

PR #559, opened 2022-12-29
Move CI to GitHub Actions

daun:
Switch CI checks to GitHub Actions.

**Goals**

- Avoid CircleCI usage limits (test runs are hidden when free quota is met)
- Speed up test runs (GitHub actions looks to be 50-70% faster than CircleCI)

**Tasks**

- [x] Create GitHub action for E2E tests with Cypress ([see workflow runs](https://github.com/swup/swup/actions/workflows/ci.yml))
- [x] Run the `build` script only once, instead of three times:
  - as part of `npm ci`
  - as a separate `npm run build`
  - as part of `npm run test`
- [x] Add back Cypress recording integration (project id and record key)
- [x] Integrate `lint` step here as well?

Solves #540


package.json
```diff
@@ -4,19 +4,20 @@
...
     "build": "webpack-cli",
     "lint": "prettier src/**/*.{js,mjs} --write",
     "prepublish": "npm run compile && npm run build",
     "postinstall": "opencollective-postinstall || true",
-    "ci": "npm run build && npm run test:instrument && start-server-and-test test:server http://localhost:8274 test:run:record",
-    "test": "npm run build && npm run test:instrument && start-server-and-test test:server http://localhost:8274 test:run",
-    "test:headed": "npm run build && npm run test:instrument && start-server-and-test test:server http://localhost:8274 cy:open",
+    "ci": "start-server-and-test test:start 8274 cy:run:record",
+    "test": "start-server-and-test test:start 8274 cy:run",
+    "test:dev": "start-server-and-test test:start 8274 cy:open",
```
gmrchk, line 14: Am I missing something, or are the port numbers duplicated and don't actually do anything here?

gmrchk (approved):

package.json
daun: They are duplicated but I can't think of a good way of sharing them. They're also hardcoded into the test suite to check the domain/origin, we're hardcoding them here as well to make sure it's always the same port on each run.

PR #560, opened 2022-12-30
Store version in swup instance

daun:
- Create `swup.version` property from package.json to allow checking the installed swup version
- Create a helper function for checking installed vs. required versions
- The helper isn't bundled into swup itself, it's just provided for plugins to import and use

Output swup version:

```js
swup.version; // '3.0.0'
```

Check for required version:

```js
import { versionSatisfies } from 'swup/helpers';

versionSatisfies('3.0.0', ['2'])              //  → true
versionSatisfies('3.0.0', ['3'])              //  → true
versionSatisfies('3.0.0', ['3.1'])            //  → false
versionSatisfies('3.4.0', ['>=3.0', '<4.0'])  //  → true
versionSatisfies('3.4.0', ['>=2.0', '<3.0'])  //  → false
```


src/index.js
```diff
@@ -12,6 +12,7 @@ import renderPage from './modules/renderPage.js';
 import triggerEvent from './modules/triggerEvent.js';
 import updateTransition from './modules/updateTransition.js';
 
+import { version } from '../package.json';
```
hirasso, line 15: This throws on my (webpack) test site, where I am importing and compiling `swup/src/index.js` directly:

```shell
Can't import the named export 'version' (imported as 'version') from default-exporting module (only default export is available)
```
I have never imported a JSON file into a script, so not much experience here. This compiles for me, though:

```js
import packageJSON from '../package.json';
// ...later in the class:
version = packageJSON.version
```

src/utils/index.js
```diff
@@ -34,3 +34,24 @@ export const escapeCssIdentifier = (ident) => {
...
+};
+
+// Numerically compare version strings after normalizing them
+export const compareVersion = (a, b) => {
+	a = normalizeVersion(a);
+	b = normalizeVersion(b);
+	return a.localeCompare(b, undefined, { numeric: true });
+};
+
+// Check if the required version is installed
```
hirasso, line 54: Hmm, not completely convinced if this will be enough in all scenarios. I would have imagined three methods: `versionHigherThen`, `versionLowerThen`, `versionEquals` or the like.

hirasso (commented):
Nice! This will come in handy. Added two comments below.

src/utils/index.js
hirasso: (wrong line selected, I mean the `checkVersion` method)

src/index.js
daun: I'll have to look into this then. Bundling locally with microbundle seems fine, it should theoretically just import the actual version as a string '3.0.0'.

daun: Your example -- does that bundle just the version string or the whole package.json as an object? In the dist file, I mean.

src/utils/index.js
daun: Like what scenarios for example? I was assuming a greater-than-or-equals check is enough for a feature check but maybe that's too simple?

src/index.js
hirasso: Only the ID:

```js
// CONCATENATED MODULE: ../swup/package.json
var package_namespaceObject = {"i8":"3.0.0"};
// ...
__publicField(this, "version", package_namespaceObject.i8);
```

`amdName` for example doesn't make it in the compiled file.

src/utils/index.js
hirasso: Basically it's just the API that confuses me. `checkVersion` is not clear enough, to me it reads like it would only return `true` for an exact version match. 

Another question, why are you proposing to have `swup.version` as the first argument? Are you planning to use this function to check for plugin versions as well sometime in the future?

hirasso: How about we remove the `checkVersion` method and only keep `compareVersion`? That function is much simpler for me to understand (predict)

hirasso: Anyways, none of that is really that important. Don't want to slow us down. Ready to merge from my side, if the above version import works reliably.

This review process needs to be learned to do well, as well. It's a bit too tempting to voice an opinion on every last thing 😅🤐

daun: Let's find a better name then. I think it's useful to add that abstraction, the compareVersion is a bit unwieldy since it requires wrangling with -1, 0 and 1.

daun: Yeah the idea was to supply a utility function that's useful in all contexts, so we can't hardcode the swup version.

hirasso: `versionEqualOrHigher` ? Longer but clearer IMO

daun: Just looked up how the semver package calls that. It's either `satisfies` for random strings or `minVersion` for specific versions. I like `minVersion` 🤠

src/index.js
daun: I've switch to your suggested way of importing the version. Let me know if that fixes it on your end.

src/utils/index.js
hirasso: Great, `minVersion` it is! Another question: How would consumers (or plugins) be able to import these functions from `utils`? My first guess was that we would need to make the function available as an instance method of `Swup`:

```js
import { minVersion } from './utils/index.js';
export default class Swup {
    // ...
    this.minVersion = minVersion;
    // ...
}
```
But probably you had something more elegant in mind?

daun: It should work like in the example above, by importing from swup's utils export:  

```js
import { minVersion } from 'swup/utils'
```

The advantage of **not** putting it on the swup instance is that we don't increase swup's bundle size. Only if any plugins need to check for a specific version, they can import the util and bundle it themselves. Util-on-demand 🤠

src/index.js
```diff
@@ -12,6 +12,7 @@ import renderPage from './modules/renderPage.js';
 import triggerEvent from './modules/triggerEvent.js';
 import updateTransition from './modules/updateTransition.js';
 
+import pckg from '../package.json';
```
gmrchk, line 15: Unless I'm mistaken, this will always bundle the whole JSON file, which we likely don't want. 
One of the workarounds for this is to use a global variable and use something like webpacks DefinePlugin to replace it with the actual value from `package.json`.

daun: I thought so as well, but interestingly tree-shaking takes care of that nowadays. If you inspect the `dist` folder after running `npm install` and `npm run build` on the PR branch, it's just the version string. Rasso checked with webpack as well, same thing.

src/utils/index.js
```diff
@@ -34,3 +34,24 @@ export const escapeCssIdentifier = (ident) => {
...
+export const compareVersion = (a, b) => {
+	a = normalizeVersion(a);
+	b = normalizeVersion(b);
+	return a.localeCompare(b, undefined, { numeric: true });
+};
+
+// Check if the required version or higher is installed
+export const minVersion = (installed, required) => {
+	return compareVersion(installed, required) >= 0;
+};
```
gmrchk, line 57: Since these are not part of the main bundle (right?), we might as well just use the [semver](https://www.npmjs.com/package/semver) package under the hood? Seem like there should be a overlap in the functionality.

src/utils/index.js
hirasso: Sounds great!

src/index.js
gmrchk: Damn, we're living in a future. 🤷 
It seems like the webpack is not on the new version tho, so I'm not sure where this update would come from. It used to be a problem as long as I can remember. 
Anyways, you checked, soooo... 👍

src/utils/index.js
daun: The semver package is quite bloated as it has to deal with lots of complex patterns and edge cases. I'm wondering if we should create this at all, actually. Let's discuss tomorrow...

src/index.js
hirasso: Actually, I didn't check it with swups webpack version, but with my own v5 webpack (importing `swup/src/index.js` directly, that's how I always test the changes)

daun: Good to know, I'll check against webpack 4 as well -- even though we won't be on webpack anymore since this is on the `next` branch.

daun:
I've updated the helper, it's now called `versionSatisfies` and it can check for multiple requirements and supports comparison operators like `>=` and `<` for giving exact requirements. See the updated example above. I think this is quite robust for use in plugins.

daun:
Tested and fixed treeshaking of the package.json import on webpack 4.

daun:
Ready for review 🤠

daun:
@hirasso I'd appreciate if you could try the new `import * as pckg from 'package.json'` syntax in your webpack 5 setup one last time to make sure. Apparently that's the only way to have treeshaking for json imports in webpack 4 and 5.

hirasso:
Hmm, with the import `import * as pckg from 'package.json'` I'm getting the error for `version = pckg.version` again:
```
ERROR in ../swup/src/index.js 33:35-47
Can't import the named export 'version' (imported as 'pckg') from default-exporting module (only default export is available)
```
`console.log(pckg)` reveals that the import is an `esModule` with a nested `default` property:
```
{
    "default": {
        "name": "swup",
        "amdName": "Swup",
        "version": "3.0.0",
        "description": "Complete, flexible, extensible and easy to use page transition library for your web.",
        "type": "module",
        "source": "src/index.js",
        "main": "./dist/index.cjs",
        "module": "./dist/index.module.js",
        "unpkg": "./dist/index.umd.js",
        "exports": {
            ".": "./dist/index.modern.js",
            "./*": "./dist/*.modern.js",
            "./lib/*": "./dist/*.modern.js"
        },
        "files": [
            "src",
            "dist"
        ],
        "scripts": {
            "cy:open": "cypress open",
            "build": "microbundle src/*.js -f modern,esm,cjs && microbundle -f umd --external none",
            "dev": "microbundle src/*.js -w",
            "lint": "prettier src/**/*.{js,mjs} --write",
            "prepublish": "npm run build",
            "postinstall": "opencollective-postinstall || true",
            "ci": "npm run build && npm run test:instrument && start-server-and-test test:server http://localhost:8274 test:run:record",
            "test": "npm run build && npm run test:instrument && start-server-and-test test:server http://localhost:8274 test:run",
            "test:nobuild": "npm run test:instrument && start-server-and-test test:server http://localhost:8274 test:run",
            "test:headed": "npm run build && npm run test:instrument && start-server-and-test test:server http://localhost:8274 cy:open",
            "test:instrument": "nyc instrument --compact=false dist test/site/swup",
            "test:server": "http-server --port 8274 test/site",
            "test:run": "cypress run",
            "test:run:record": "cypress run --record"
        },
        "author": "Georgy Marchuk",
        "license": "MIT",
        "repository": {
            "type": "git",
            "url": "https://github.com/gmrchk/swup.git"
        },
        "keywords": [
            "css",
            "animation",
            "page",
            "transition"
        ],
        "dependencies": {
            "delegate-it": "^3.0.1",
            "opencollective-postinstall": "^2.0.2"
        },
        "devDependencies": {
            "@cypress/code-coverage": "^3.10.0",
            "@swup/prettier-config": "^1.0.0",
            "cypress": "^10.0.2",
            "http-server": "^14.1.1",
            "istanbul-lib-coverage": "^3.2.0",
            "microbundle": "^0.15.0",
            "nyc": "^15.1.0",
            "prettier": "1.17.0",
            "start-server-and-test": "^1.14.0"
        },
        "collective": {
            "type": "opencollective",
            "url": "https://opencollective.com/swup"
        },
        "prettier": "@swup/prettier-config"
    }
}
```
Changing the import to the following compiles and tree-shakes as intended (only the `version` ends up in the build):
```js
import { default as pckg } from '../package.json';
```
Reminds me of an [issue I once experienced with VITE](https://github.com/vitejs/vite/issues/5499)

So now's the question: Does `import { default as pckg } from '../package.json';` work on WebPack 4? 🌀

daun:
Are you sure you're on the latest branch? That package.json looks out of date and we're no longer importing a named export. Does the first line in swup's `index.js` say `import * as pckg from '../package.json'`? Just to confirm.

daun:
@hirasso Are you sure you're on the latest branch? That package.json looks out of date and we're no longer importing a named export. Does the first line in swup's `index.js` say `import * as pckg from '../package.json'`? Just to confirm.

hirasso:
Yes, the branch is up-to-date. If you want, we could jump on a quick call tomorrow morning to debug this together.

daun:
@hirasso Happy to get on a call! This is bugging me, haha.

daun:
I've fixed the webpack issues, made sure it's compiling on v4 and v5, added a test for the version being present on the instance and cleaned everything up a bit. I'll just merge this as it's tested and all checks went through 🧃

hirasso:
Just to report back: Works perfectly now with my WebPack5 setup! Also `versionSatisfies` feels very intuitive, nice work!

PR #562, opened 2023-01-01
Refactor event handling

daun:
- Simplify removal of event handlers using Array methods
- Will decrease bundle size by a few bytes
- Added test to ensure it's still working correctly


hirasso (approved):
Tests are passing 🥁

Issue #561, opened 2022-12-30
Allow plugins to delay start of in-animation


daun:
I've been trying to get the head plugin to wait for new stylesheets to finish loading before starting the in-animation. Sadly, there's currently no way of manually deferring the in-animation until a Promise resolves. [Here's what I've tried so far](https://github.com/swup/head-plugin/compare/master...await-assets).

Not sure how the API for such a thing would look like. Turbolinks has a nice `pause` and `resume` API working with async methods for all events.

```js
document.addEventListener('turbo:before-render', async (event) => {
  event.preventDefault()
  await animateOut()
  event.detail.resume()
})
```

**Update after some research**

The issue seems to be that `enterPage` will remove the `is-animating` class immediately, thus starting the out-animation and allowing no way of cancelling or extending it by adding a Promise.

https://github.com/swup/swup/blob/8b7dd1866859f22b2e94e352e0bc33d4b918ab90/src/modules/enterPage.js#L9-L14

**Likely solution**

Allow pausing execution after the `willReplaceContent` event and before the actual replacing of the content. Plugins can then defer rendering new content before stylesheets are loaded.

daun:
Solved in #567

PR #565, opened 2023-01-02
Move path classnames to route-name plugin

daun:
**Description**

- Proposal to stop adding `to-{url}` classnames by default in swup
- Those classnames should and could be optional
- The [route-name](https://github.com/swup/route-name-plugin) plugin was [just updated](https://github.com/swup/route-name-plugin/pull/2) to start adding those classnames, fitting its purpose
- We'll slim down the core while providing a clear alternative for people migrating to swup 3

**Before: swup only**

```html
<html class="is-animating to-b">
```

**After: swup only**

```html
<html class="is-animating">
```

**After: swup with route-name plugin**

```html
<html class="is-animating from-a to-b">
```

**Upgrade path**

```js
const swup = new Swup({
  plugins: [new SwupRouteNamePlugin({ paths: true })]
})
```

**Checks**


hirasso:
If I'm not mistaken, JS Plugin expects the route name classes to be added by core. If we extract the functionality, we would create plugin peer dependencies.

daun:
@hirasso Could you point me to where it expects those classes to be set? I haven't used the js plugin myself so I'm not too familiar with it.

hirasso:
@daun If you look at the documentation of JS Plugin:

```js
const options = [
  {
    from: '(.*)', // matches any route
    to: '(.*)', // matches any route
    out: next => next(), // immediately continues
    in: next => next() // immediately continues
  }
];
```
It's an array of `from/to` directives. You can use any route in the `from` and `to` properties of each directive.

Actually, JS Plugin would also help with waiting for any transition to start/finish (you have written about that in #561), so probably worth checking it out :)

daun:
@hirasso I see, seems like that's not related then. The JS plugin does its own parsing of the URL into route names. It doesn't add or require classnames on the `html` element. All this PR does is make swup stop adding the `to-{url}` classnames to the HTML and move that feature to the route-name plugin instead. There's some overlap between the js and route-name plugin in that they both use path-to-regexp to parse urls into named routes, but that's coincidental.

Thanks for the pointer about the delay issue! I'll need to see if that helps with that.

hirasso:
@daun great, feels like we are gaining a broader overview of swup+plugins step by step 🐾

hirasso (approved):
Looks great, lets gooo

PR #567, opened 2023-01-03
Allow defering in-animation

daun:
**Description**

Allow plugins to defer the in-animation by:

- Move the logic for replacing blocks and document title into separate `replaceContent` method
- Make `replaceContent` return a Promise and wait for it before starting the in-animation
- This allows overwriting the method and returning a custom or a chained Promise, pausing execution
- No breaking changes as all external APIs and existing methods are untouched

**Solves**

#561 

**Checks**

### DIFF
--- a/src/modules/replaceContent.js
+++ b/src/modules/replaceContent.js
@@ -1,0 +1,17 @@
+import { query } from '../utils.js';
+
+const replaceContent = function({ blocks, title }) {
+	// replace blocks
+	blocks.forEach((html, i) => {
+		const block = query(`[data-swup="${i}"]`, document.body);
+		block.outerHTML = html;
+	});
+
+		// set title
+	document.title = title;
+
+	// Return a Promise to allow plugins to defer
+	return Promise.resolve();
+};
+
+export default replaceContent;
--- a/src/modules/replaceContent.js
+++ b/src/modules/replaceContent.js
@@ -1,5 +1,3 @@
-import { query } from '../utils.js';
-
 /**
  * Perform the replacement of content after loading a page.
  *
@@ -15,7 +13,7 @@ import { query } from '../utils.js';
 const replaceContent = function({ blocks, title }) {
 	// Replace content blocks
 	blocks.forEach((html, i) => {
-		const block = query(`[data-swup="${i}"]`, document.body);
+		const block = document.body.querySelector(`[data-swup="${i}"]`);
 		block.outerHTML = html;
 	});
 
--- a/src/modules/replaceContent.js
+++ b/src/modules/replaceContent.js
@@ -1,13 +1,25 @@
 import { query } from '../utils.js';
 
+/**
+ * Perform the replacement of content after loading a page.
+ *
+ * This method can be replaced or augmented by plugins to allow pausing.
+ *
+ * It takes an object with the page data as return from `getPageData` and has to
+ * return a Promise that resolves once all content has been replaced and the
+ * site is ready to start animating in the new page.
+ *
+ * @param {object} page The page object
+ * @returns Promise
+ */
 const replaceContent = function({ blocks, title }) {
-	// replace blocks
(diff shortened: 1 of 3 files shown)


src/modules/replaceContent.js
```diff
@@ -0,0 +1,29 @@
...
+ * return a Promise that resolves once all content has been replaced and the
+ * site is ready to start animating in the new page.
+ *
+ * @param {object} page The page object
+ * @returns Promise
+ */
+const replaceContent = function({ blocks, title }) {
+	// Replace content blocks
+	blocks.forEach((html, i) => {
+		const block = query(`[data-swup="${i}"]`, document.body);
```
hirasso, line 18: I would switch from `query` to `document.querySelector` here. One less import/dependency from swup...

src/modules/replaceContent.js
```diff
@@ -0,0 +1,29 @@
...
+ * This method can be replaced or augmented by plugins to allow pausing.
+ *
+ * It takes an object with the page data as return from `getPageData` and has to
+ * return a Promise that resolves once all content has been replaced and the
+ * site is ready to start animating in the new page.
+ *
+ * @param {object} page The page object
+ * @returns Promise
+ */
+const replaceContent = function({ blocks, title }) {
```
hirasso, line 15: Could be an arrow function. But I don't care that much tbh :)

hirasso:
Just tested this, works as intended:

```js
const swup = new Swup(options);
swup.replaceContent = ({ blocks, title }) => {
	// Replace content blocks
	blocks.forEach((html, i) => {
		const block = document.querySelector(`[data-swup="${i}"]`);
		block.outerHTML = html;
	});

	// Update browser title
	document.title = title;

	// Return a Promise to allow plugins to defer
	return new Promise((resolve) => {
    setTimeout(resolve, 5000);
  });
};
```

This will pause swup for 5 seconds after replacing the content.

I'm still not completely sure what kind of use cases the returned promise would enable. Do you have a use case where this would be necessary?

Making the `replaceContent` function `pluggable` could be very powerful, though.

src/modules/replaceContent.js
daun: I think it needs to be a normal function so we can stick it on the swup instance and it keeps its `this` reference. Not import here, anyway.

hirasso: Interesting... just tested it, you are right! With an arrow function, `this` becomes undefined.

daun:
@hirasso The main use case is pausing before the replacement happens, not after. That allows waiting for stylesheets of the new page to load before placing the new HTML.

Not tested, could break horribly:

```js
swup._originalReplaceContent = swup.replaceContent;

swup.replaceContent = (...args) => {
  const stylesheetLoadedPromise = new Promise ((resolve) => setTimeout(resolve, 5000));
  return new Promise((resolve) => {
    stylesheetLoadedPromise.then(() => {
      swup._originalReplaceContent(...args).then(resolve);
    });
  });
};
```

hirasso:
😳 what is that wizardry?! 😄 But ok, I forgot about the stylesheets. So you intend to use it in the head plugin?

hirasso (approved):

daun:
> 😳 what is that wizardry?! 😄 But ok, I forgot about the stylesheets. So you intend to use it in the head plugin?

Have you tried the code? Does it work on your end? And yeah, it's for the head plugin. I like splitting up stylesheets on large sites and it currently doesn't work too well with swup.

hirasso:
The code above works, yes. Haven't tried it with actual dynamically loaded stylesheets for different pages, I never do this.

daun:
@hirasso Perfect :) The dynamic stylesheets only make sense if something like the head plugin is able to automate the process. I don't want to be doing that myself either. Some projects just get so large with so many subpages that loading the whole stylesheet on first visit would slow things down a bit.

Issue #566, opened 2023-01-03
Switch to fetch API


daun:

hirasso:
Yes, let's do this! And let's make it abortable while we're at it 🏝️

daun:
Closed via #692

PR #569, opened 2023-01-03
Switch to fetch API

daun:
**Description**

- [x] Switch to native `fetch` API
- [x] Add a compatibility layer for methods and plugins expecting XHR properties like `responseURL`
- [x]  Allow aborting requests
- [x] Test if server errors (500) are propagated properly as before
- [ ] Test this with plugins (forms, preload and custom-payload plugins specifically)
    - [ ] Make sure none of the plugins depend on the previously returned `xmlHttpRequest` directly

**Checks**

**Closes**

#566

### DIFF
--- a/src/helpers/fetch.js
+++ b/src/helpers/fetch.js
@@ -1,34 +1,22 @@
-const fetch = (setOptions, callback = false) => {
-	let defaults = {
-		url: window.location.pathname + window.location.search,
-		method: 'GET',
-		data: null,
-		headers: {}
-	};
-
-	let options = {
-		...defaults,
-		...setOptions
-	};
-
-	let request = new XMLHttpRequest();
-
-	request.onreadystatechange = function() {
-		if (request.readyState === 4) {
-			if (request.status !== 500) {
-				callback(request);
-			} else {
-				callback(request);
-			}
-		}
-	};
-
-	request.open(options.method, options.url, true);
-	Object.keys(options.headers).forEach((key) => {
-		request.setRequestHeader(key, options.headers[key]);
-	});
-	request.send(options.data);
-	return request;
+const defaults = {
+	url: window.location.pathname + window.location.search,
+	method: 'GET',
+	headers: {}
 };
 
-export default fetch;
+const fetchWrapper = (options, callback = false) => {
+	const { url, method, data, body, headers } = { ...defaults, ...options };
+
+	const init = { method, headers, body: body || data };
+
+	return fetch(url, init)
+		.then((response) => response.text())
+		.then((html) => {
+			// Compatibility layer for other methods expecting XHR properties
+			response.responseURL = response.url;
+			response.responseText = html;
+			callback(response);
+		});
+};
+
+export default fetchWrapper;
--- a/src/helpers/fetch.js
+++ b/src/helpers/fetch.js
@@ -1,11 +1,14 @@
-const defaults = {
-	url: window.location.pathname + window.location.search,
-	method: 'GET',
-	headers: {}
-};
-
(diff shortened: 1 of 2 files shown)


hirasso:
@daun I just had time to give the aborting of the request a try – hope it's ok that I'm joining you in this PR

hirasso:
...just tested error detection and reload for 500 errors, still works as expected 👍

src/helpers/fetch.js
```diff
@@ -1,34 +1,45 @@
...
+			callback(response);
+		})
+		.catch((error) => {
+			// Handle intentionally aborted requests
+			if (signal.aborted && signal.reason === 'aborted') {
+				callback({ error: 'aborted' });
+				return;
 			}
-		}
+			// Handle all other errors
```
hirasso, line 34: I just assumed that `fetch` can error for other reasons as well (e.g. network timeout?) That's why I added the callback here as well.

src/modules/loadPage.js
```diff
@@ -37,24 +37,38 @@ const loadPage = function(data, popstate = false) {
...
+			const { abort } = fetch(
+				{ ...data, headers: this.options.requestHeaders },
+				(response) => {
+					// Reject without URL for aborted requests (will prevent browser reloads)
+					if (response.error === 'aborted') {
+						reject();
+						return;
+					}
+					// Reject with URL for arbitary errors in fetch
+					if (response.error) {
```
hirasso, line 49: Same as inside `fetch.catch`

daun:
Nice! I do worry a bit that we're just adding bloat here? Hehe.

Let's make sure we're gaining something for the added complexity and bundle size. One idea would be extracting a `fetchPage` module that can be overwritten by the preload plugin, so that ideally swup wouldn't have to know about the concept of preloading at all. That's another PR though, let's just make sure that our changes here facilitate that kind of flexibility.

hirasso:
Swup actually _has_ to know about the concept preloading because it wants to be able to highjack a currently running preload if the user clicks a link. At least that's how far my vision reaches ;)

hirasso:
> I do worry a bit that we're just adding bloat here?

If you mean the abort controller, it's actually much nicer and cleaner to work with this than with `xmlHttpRequest.abort`. The latter:

- will always throw an error in the dev tools network tab, without a way of suppressing it
- makes it hard to detect reliably in the code that the request was aborted

daun:
By extracting a `fetchPage` method, we're giving plugins the opportunity to supply whatever promise they want from it. Which means the preload plugin can just overwrite that method and hand over a) a preloading promise or b) whatever promise the original `fetchPage` would return by default. I'll see if I can make sense of it. Maybe swup v4 🤡

daun:
> If you mean the abort controller, it's actually much nicer and cleaner to work with this than with `xmlHttpRequest.abort`

That's what I meant. If we're actually gaining readability and nicer control flow, I'm happy :)

hirasso:
> - `xmlHttpRequest.abort()` will always throw an error in the dev tools network tab, without a way of suppressing it

Actually, I have to correct myself – this is also true for aborting `fetch` with `AbortController.abort()`. 

We would have to actually implement the logic for cleanly handling `xhr.abort()` before we could say if it's more bloat then with `fetch` and `AbortController.abort`. Let's discuss if we should give `xhr` another chance – the `fetch` approach in this PR is more code then before already, even without the `abort` functionality.

hirasso:
For reference: 

https://stackoverflow.com/a/40877968/586823

daun:
Closing this in favor of #692 as a new attempt with hooks in mind.

PR #576, opened 2023-01-08
Implement plugin version check on mount

daun:
- Invoke plugin's requirement check method
- Refactor plugin module a bit
- Add tests for mounting and unmounting plugins

### DIFF
--- a/src/modules/plugins.js
+++ b/src/modules/plugins.js
@@ -6,7 +6,12 @@ export const use = function(plugin) {
 
 	this.plugins.push(plugin);
 	plugin.swup = this;
-	if (typeof plugin._beforeMount === 'function') {
+	if (plugin._checkRequirements) {
+		if (!plugin._checkRequirements()) {
+			return;
+		}
+	}
+	if (plugin._beforeMount) {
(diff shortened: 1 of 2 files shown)


src/modules/plugins.js
```diff
@@ -6,7 +6,12 @@ export const use = function(plugin) {
 
 	this.plugins.push(plugin);
 	plugin.swup = this;
-	if (typeof plugin._beforeMount === 'function') {
+	if (plugin._checkRequirements) {
```
hirasso, line 8: I think it could be helpful if the plugin could provide a reason why the requirements aren't met. This reason could then be `warn`ed to the console by swup, to give the consumers a hint. E.g. "This version of ScrollPlugin requires at least Swup v3.0.0" WDYT?

daun: Forgot to mention that the upcoming plugin base will do that part:

https://github.com/swup/plugin/blob/a4716ce5feb1279f613f650d1f7329e2cc12ad96/src/index.js#L32-40

hirasso: Ah, nice!

PR #577, opened 2023-01-09
Typescript Support

gmrchk:
**Description**

Adding TS support.
1. commit just renaming files. 
2. commit TS installation and all
3. ...
4. commit - deafcf028bb41a776c5f00f08769e392f31c91e7 actual changes to the codebase. I tried to not change the logic at all and just introduce types, but somewhere it was still needed. Few places marked in the PR. 
5. ...

* I avoided creating specific file for types, and kept it with the implementation. 
* Most functions have a return type specified. Not really needed, but felt safer when adjusting everything. 

**Checks**


src/helpers/Location.ts
```diff
@@ -8,35 +8,37 @@
...
-		Object.defineProperty(location, 'url', { get: () => location.pathname + location.search });
-		return location;
+export default class Location extends URL {
+	constructor(url: string, base: string = document.baseURI) {
+		super(url.toString(), base);
+	}
+
+	get url() {
+		return this.pathname + this.search;
 	}
```
gmrchk, line 24: Pretty much only place with updated logic, since couldn't figure out what `defineProperty` does to the original one.

src/modules/leavePage.ts
```diff
@@ -13,7 +22,7 @@ const leavePage = function(data, { popstate = false, skipTransition = false } =
 	}
 
 	// animation promise stuff
-	const animationPromises = this.getAnimationPromises('out');
+	const animationPromises: Promise<void>[] = this.getAnimationPromises();
```
gmrchk, line 25: This is one of the places where code was actually updated, since TS showed the param was never used.

daun:
This is great 🧃 Happy to have this part of the upcoming release.

src/modules/leavePage.ts
daun: The `in` and `out` string arguments should definitely be part of it. Swup itself doesn't use them but the JS plugin expects them to be there I think?

src/helpers/Location.ts
```diff
@@ -8,35 +8,37 @@
...
 	/**
 	 * Instantiate a Location from an element's href attribute
 	 * @param {Element} el
 	 * @return new Location instance
 	 */
-	static fromElement(el) {
+	static fromElement(el: HTMLAnchorElement): Location {
 		const href = el.getAttribute('href') || el.getAttribute('xlink:href');
-		return new Location(href);
+		return new Location(href!);
```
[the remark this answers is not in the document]
gmrchk: Since we're getting the value with `getAttribute`, it might be `null`. `!` basically tells TS that we know it will be of some value (in this case `string`). 
This is not the best and it would be better to throw error if something like that happens, because it shouldn't, but that would be changing the implementation, so I avoided that.

src/helpers/Location.ts
gmrchk: Aha, I see! I was wondering about that.  
Hm that's a tough one... Can we setup the target browsers in a way that we won't need the polyfil (and babel won't add it), or is it still required?

daun: Yes, fiddling with the babel target is the way to go here. I'll give it a try!

src/modules/leavePage.ts
gmrchk: Oh man, that must be it... I'll update it.

gmrchk: Another great example how overwriting the Swup instance can be dangerous. 😄

src/helpers/Location.ts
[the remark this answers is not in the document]
daun: Sounds good!

src/helpers/Location.ts
daun: I've tested a browserslist config that gets rid of 1K of polyfills by skipping Safari 12 🤡 WIll create a PR.

src/modules/leavePage.ts
daun: Haha, yeah, I've also pushed commits in the past removing that same argument but somehow remembered before merging...

daun:
Tested in a few projects and working like a charm with loads of plugins. Good to go from my side, maybe @hirasso could give it a last try locally as well before we merge and ship.

There was some confusion for tsc about the root dir vs. the output dir since we're technically in `./src` but importing `package.json` from the `root`. Should be good now. My IDE shows all types for the main as well as the named sub-exports.

hirasso:
I also just tested this branch with the website I'm currently working on, no errors 💯

PR #581, opened 2023-01-10
Make swup agnostic of preload plugin

daun:
**Description**

- Extract `fetchPage` method that returns a Promise for loading a page
- Allow plugins to overwrite that method, optionally returning a custom promise
- Preload plugin will overwrite method and store promise on itself (see [PR #56](https://github.com/swup/preload-plugin/pull/56))

**Checks**

- [x] The PR is submitted to the `typescript` branch

### DIFF
--- a/src/modules/loadPage.ts
+++ b/src/modules/loadPage.ts
@@ -1,6 +1,5 @@
 import { classify, createHistoryRecord, getCurrentUrl } from '../helpers.js';
 import Swup from '../index.js';
-import { PageRecord } from './Cache.js';
 
 export type TransitionOptions = {
 	url: string;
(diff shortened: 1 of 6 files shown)


src/modules/loadPage.ts
```diff
@@ -33,41 +29,10 @@ const loadPage = function(this: Swup, data: TransitionOptions, popstate: PopStat
...
-				}
-				// render page
-				const cacheablePageData = { ...page, url };
-				this.cache.cacheUrl(cacheablePageData);
-				this.triggerEvent('pageLoaded');
-				resolve(cacheablePageData);
-			});
-		});
-	}
+	const fetchPromise = this.fetchPage(data);
```
hirasso, line 32: Wouldn't we need a way to take over an already-running fetch promise here? `swup. getPageFetchPromise(url)` seemed like a very good idea to me, where preload plugin could add fetch promises to an object `swup.fetchPromises` and swup could pick those up on demand.

PR #584, opened 2023-01-11
Exports from the Swup package

gmrchk:
**Description**

This should make it possible to import additional named exports from `swup` instead of importing from nested paths like `swup/helpers` like we used to before. 

```js
import Swup, { getCurrentUrl } from 'swup';
```

It should also make the Swups API more stable, since any breaking change to the `index.ts` is now  breaking change, instead of possibly making breaking changes by updating random files (happened before that we broke some plugins). 

**Checks**


daun:
@gmrchk I've added a few fixes and tested in webpack 4 and 5, compiling fine. Good to merge!

daun (approved):

PR #588, opened 2023-01-15
Use Native git hooks

hirasso:
Alternative to #587 

**Description**

Instead of relying on a library like husky, we could just use the built-in `git config core.hooksPath` directive to set up the git hooks.

I just tested it with `npm install`, the git hooks directory is correctly set to `.githooks`. Nice and simple.


hirasso:
For auto-formatting instead of only linting staged files, this package looks like the way to go:

https://www.npmjs.com/package/git-format-staged

It requires Python, though, which is only pre-installed on OSX and Linux. [(Source)](https://prettier.io/docs/en/precommit.html#option-5-git-format-stagedhttpsgithubcomhallettjgit-format-staged)

daun:
I'm getting an error when commiting with this setup:

```
.githooks/pre-commit: line 3: npm: command not found
```

Not sure if this is a setup issue as npm is definitely installed and I can run `npm run lint` just fine.

daun:
Does Git need to know where the `npm` binary is located? Probably not, as we're only executing the `pre-commit` file on the bash shell...

hirasso:
This could be an [IDE problem](https://stackoverflow.com/a/58539620/586823). Do you also get the error when commiting from the command line?

hirasso:
@gmrchk this is working now for @daun and me. Could you check on your end as well?

daun:
@hirasso I've pushed a solution for checking if any `*.ts` have been staged, and skipping the lint script if not. Let me know if that works for you as well.

hirasso:
@daun works like a charm! 👏

daun (approved):

hirasso:
@gmrchk I'll wait for your test results before merging

.githooks/pre-commit
```diff
@@ -0,0 +1,14 @@
...
+
+# Make sure /usr/local/bin is in the path
+PATH="$PATH:/usr/local/bin"
+
+# Check if there are any staged files ending with `.ts`
+STAGED_FILES=$(git diff --cached --name-only --diff-filter=ACM | grep "\.ts$")
+
+if [[ "$STAGED_FILES" = "" ]]; then
+  exit 0
+fi
```
gmrchk, line 11: This is way too specific for git hook - separating concerns. Hook should just run some commands at different stages, not care about what the command will do. 

As an example, once there are other types of files to lint, we'll surely update the lint command, but this will likely be forgotten meaning the lint won't run if there are no TS files in the commit. 
IMO we should just let lint command run always.

.githooks/pre-commit
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+# Make sure /usr/local/bin is in the path
+PATH="$PATH:/usr/local/bin"
```
gmrchk, line 4: This seems like something needed for possible other future hooks as well, and could be needed only for some people, meaning we could miss it in new hook files. 
Can it be extracted?

.githooks/pre-commit
daun: Sounds reasonable, let's rip it out again.

.githooks/pre-commit
daun: Not sure how. That's probably the advantage of something like husky (if they do that by default).

hirasso: @daun did the PR with Husky work out of the box for you? If yes, I would propose that we just go with that PR instead of this one.

PR #595, opened 2023-01-19
Fix ignore visit param

daun:
**Description**

Fix `ignoreVisit` param and remove origin from URL since it's never required for swup visits. Keeps query params and hash.

I'll update the docs when this is merged.

Before:

```js
ignoreVisit: (href, { el } = {}) => pattern.test(href.replace(window.location.origin, '')))
```

After:

```js
ignoreVisit: (url, { el } = {}) => pattern.test(url))
```

**Checks**

**Additional information**

### DIFF
--- a/src/__test__/index.test.ts
+++ b/src/__test__/index.test.ts
@@ -4,6 +4,8 @@ console.log = jest.fn();
 console.warn = jest.fn();
 console.error = jest.fn();
 
+const baseUrl = window.location.origin;
+
 describe('exports', () => {
 	it('exports Swup, and Options/Plugin types', () => {
 		class SwupPlugin implements Plugin {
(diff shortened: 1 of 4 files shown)


src/__test__/index.test.ts
```diff
@@ -33,4 +35,18 @@ describe('exports', () => {
...
 		expect(swup.version).not.toBeUndefined();
 	});
+
+	it('passes relative URL to ignoreVisit', () => {
+		let ignorableUrl = 'nothing';
+		const swup = new Swup({
+			ignoreVisit: (url, { el } = {}) => {
+				ignorableUrl = url;
+				return false;
+			}
```
gmrchk, line 45: Can we make the `ignoreVisit` option a `jest.fn()` and check what is it being called with instead? Seems cleaner.

src/__test__/index.test.ts
```diff
@@ -33,4 +35,18 @@ describe('exports', () => {
...
+		});
+
+		swup.shouldIgnoreVisit(baseUrl + '/path/?query#hash');
```
gmrchk, line 48: Just an idea. Feels like calling `swup.loadPage` is what really matters when it comes to ignoring navigation. Could we add a test that `shouldIgnoreVisit` is even being called when the `loadPage` is hit, and that it prevents it? e2e test is there too tho, so not important and feel free to skip.

gmrchk (approved):

daun:
@gmrchk I'll merge and release this now, and address the test suggestions in another PR. Just to minimize the time people are installing the current version without the fix.

PR #597, opened 2023-01-20
Ignore visits from loadPage method

daun:
**Description**

- Also allow ignoring visits when directly calling `swup.loadPage({ url })`
- If a visit is ignored in `loadPage`, it will force a browser visit via `window.location` (see question below)
- Make the second `popstate` parameter to `loadPage` optional for TS end users as it's mostly internal
- Add a `skipChecks` option to `loadPage` to ensure the `ignoreVisit` option is only called once, e.g. after already checking in the click and popstate handler because their logic requires the check earlier
- Rewrite unit tests with mock functions
- This would be a fix release as it was just an oversight on our part

**Update**

- Added an internal `performPageLoad` method that does not check for ignorable visits
- The public `loadPage` as well as click and popstate handlers will now use the new internal method

Also, some refactoring while at it:
- Added a new `shouldSkipTransition` method to have the logic for that live in a more obvious place
  - Incidentally, that will be useful for the upcoming fragment plugin to disable animations for fragment visits
- Instead of passing around a `popstate` argument, pass around a more generic `event` object
  - Makes sense since the argument to `triggerEvent` is called `originalEvent` and supposed to be generic

**Question**

If `loadPage` tries to load a URL and it learns that it should be ignored, it will hand over to the browser via `window.location`. That seems to make the most sense to me — ignoring a visit from a click event also lets the browser handle the visit. Skipping a popstate event also just let's it through to the browser. The alternative would be doing nothing, in which case `ignoreVisit` would have differing effects across the API. WDYT?

**Checks**


src/modules/loadPage.ts
```diff
@@ -5,17 +5,21 @@ import { PageRecord } from './Cache';
...
-) {
-	const { url, customTransition } = data;
+export const loadPage = function (this: Swup, data: TransitionOptions, popstate?: PopStateEvent) {
+	const { url, customTransition, skipChecks = false } = data;
+
+	// Check if the visit should be ignored
+	if (!skipChecks && this.shouldIgnoreVisit(url)) {
+		window.location.href = url;
+		return;
+	}
```
gmrchk, line 18: I really struggle with the idea of doing the check conditionally here the second time in the process. Since the problem is `swup.loadPage` provided on Swup to the consumer, how about we introduce a wrapper around this method with this name. Basically, it wouldn't be provided on the instance directly in this form. 
The wrapper function would be the actual accessible `loadPage` and consist of the checks, later calling this method, just like the click and popstate handlers do?

daun: Yeah, I've been thinking the same thing. How could we call the internal API method that wouldn't do the check -- performPageLoad? executePageLoad?

daun: That would also let us get rid of the internal popsate argument on the external method.

gmrchk:
Didn't give it too much thought, but...
I think the default behaviour of `loadPage` when the URL is ignored is not wrong now, when it does nothing. After all, that's exactly what the click handler will do as well, just do nothing and don't cancel the default event for the browser either.

On the other hand, I understand that calling `loadPage` already indicates you'd like to navigate to some URL. If it's ignored by Swup, the usual navigation should happen obviously.

It's hard to imagine the use case for this tbh, so hard to decide also. Let's go with the one you introduced already, eg. doing native navigation as fallback. 
Does it need to be documented? 🤔

src/modules/loadPage.ts
gmrchk: Both will do, it doesn't matter much since it can be always changed. Maybe `performNavigation` or `startTransition`?

hirasso:
Are we certain that `ignoreVisit` is actually really the same thing as `dontLoadPage` in every last use case? In my view `loadPage` should always be executed, just as it is right now in `master`.  Maybe easier to talk this one through?

daun:
I would argue that `loadPage()` is just another way of saying `visit()`, so my expectation would definitely be that swup will check if this visit applies to it, and if not, fall back to default browser behavior. Same as everywhere else: click, popstate, form submission, etc.

hirasso:
Fair enough.

daun:
I've pushed a few updates addressed by @gmrchk that should get rid of the double/conditional check. And some other optimizations and streamlining around the page-load lifecycle. See notes under heading *Updates* above.

src/modules/renderPage.ts
```diff
@@ -2,12 +2,15 @@ import { Location, updateHistoryRecord, getCurrentUrl } from '../helpers';
 import Swup from '../Swup';
 import { PageRecord } from './Cache';
 
+export type PageRenderOptions = {
+	event?: Event;
```
gmrchk, line 6: `PopStateEvent` seems like a more exact option here. Could you explain why the generalization of the type here?

src/Swup.ts
```diff
@@ -95,6 +95,7 @@ export default class Swup {
 	boundPopStateHandler: (event: PopStateEvent) => void;
 
 	loadPage = loadPage;
+	performPageLoad = performPageLoad;
 	leavePage = leavePage;
 	renderPage = renderPage;
 	replaceContent = replaceContent;
```
gmrchk, line 101: Up until now, everything was more or less public, and we're introducing private methods. We could specify that it's private so it's reflected in generated types as well, but that will likely cause problems in our methods files too when using these unfortunately. We should figure out how to make private things private for the outside consumers.

gmrchk: Well ok, it's actually mostly private and exposed as public, so the problem is there for some time, but still, anything we could do? 🤔

src/modules/renderPage.ts
daun: While rethinking the types for the `loadPage` method without a `popstate` param, I thought about how useful it would be to have a reference to **any** type of event that triggered a visit. Right now it's just popstate events, but I can see lots of valid use cases for having access to the click and formsubmit events as well and passing them along to swup event handlers. That kind of matches the intention of the `triggerEvent` method, that passes along an `originalEvent` and not a `popstateEvent`. Does that make sense?

src/Swup.ts
daun: True, right now they're kind of half public, half private/undocumented. While we could start adding underscore like `_performPageLoad`, we should do that for all undocumented ones like `renderPage` and `leavePage` as well. Maybe that's a breaking change for v4? Browser support for [private methods](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields) isn't very good at the moment either.

src/modules/renderPage.ts
gmrchk: Not sure about it. The consumer will always have to cast the event in handler to `as PopStateEvent` for example, to be able to read that it has a `state` on it, but I guess that's the case even now since multiple events are possible.  
Anyways, let's go with it! 👍

src/Swup.ts
gmrchk: Basically, it's a mess now. Let's address later. 😄 
No need for renaming, TypeScript has the support, and as long as it's reflected in types for consumers, I think that's ok.

gmrchk: https://github.com/swup/swup/issues/598

gmrchk (approved):

src/modules/renderPage.ts
daun: True, right now it's a mouse or a popstate event, so the casting is already necessary.

src/Swup.ts
daun: Perfect, let's brainstorm solutions in the issue.

Issue #602, opened 2023-01-23
Garbage Collector for `swup.cache`


hirasso:
**Describe the problem**

On very large websites the cache can become a problem on less-powerful end devices. It's basically be a built-in memory-leak. I would propose to introduce a configurable garbage collector.

**Describe the proposed solution**

- [x] #603 
- [ ] ~Every time a new page is added to the cache, clean up old pages~
- [ ] ~The amount of pages allowed in the cache should be configurable through an option~
- [ ] ~There should be a way for pages to be ignored by the garbage collector (maybe a callback that will set a property `persist` in `PageHtmlData`?)~

The above mechanism won't be implemented, instead see #607

gmrchk:
I've never had any issues with this, and there are much much larger applications that run in browsers nowadays. On the other hand, I've always had a "beast" of a machine, so my experience is probably not the one of a usual user. 

Since this is not something that would always be needed, it could probably be implemented in a form of a plugin, which would just access the cache every now and then (we can brainstorm the logic here) and remove the oldest records. What we're missing in the cache is the last update timestamp tho, which needs to be added to Swup.

daun:
I've also never found this to be an issue across many swup projects, but I understand that it can become one under certain circumstances. It's an array of very simple objects with strings and no DOM elements in them, so I wouldn't worry about the memory part but I get the concern.

What I'd find more useful is purging the cache after a certain amount of time to avoid stale data. Making this a plugin — Cache++ with some sort of filtering or callback — would feel right to me, in terms of scope and complexity.  Is anything required from the core side to make that happen? It can already hook into `pageCached` to add the timestamp, and into `pageVisit` to purge the entries.

daun:
Switching to a `Map` for the cache entries would make sense to allow easier iteration, slicing, sorting, etc.

hirasso:
Actually, thinking about this further, I would prefer to introduce a very slim timestamp-based  cache invalidation mechanism in swups core instead of a separate plugin. I think an increase in package size is ok where it really makes sense. Or what would be the advantage of having this in a separate plugin, apart from core bundle size, @daun ?

daun:
Conceptually speaking, given how almost everything in swup is a plugin, the whole cache **could** be a plugin. It's debatable if it **should** be a plugin since 90% of all projects would install a cache plugin anyway, but architecturally it'd make sense, I guess. I don't think we should rip out the cache since it's very useful and popular, but I think we shouldn't let it get much more complex.

[...]

PR #603, opened 2023-01-23
Switch to `Map` for `cache.pages`

hirasso:
**Description**

Use a `Map` instead of a `Record` for `cache.pages`, as proposed in https://github.com/swup/swup/issues/602#issuecomment-1400248073

**Checks**

- [x] The PR is submitted to the ~`master`~ `next` branch, since it's a breaking change. Consumers might be directly accessing the `pages` object in their code bases.

### DIFF
--- a/src/modules/fetchPage.ts
+++ b/src/modules/fetchPage.ts
@@ -9,7 +9,7 @@ export function fetchPage(this: Swup, data: TransitionOptions): Promise<PageReco
 
 	if (this.cache.exists(url)) {
 		this.triggerEvent('pageRetrievedFromCache');
-		return Promise.resolve(this.cache.getPage(url));
+		return Promise.resolve(this.cache.getPage(url) as PageRecord);
 	}
 
 	return new Promise((resolve, reject) => {
--- a/src/modules/Cache.ts
+++ b/src/modules/Cache.ts
@@ -7,7 +7,7 @@ export interface PageRecord extends PageData {
 	responseURL: string;
 }
 export class Cache {
-	pages: Record<string, PageRecord> = {};
+	pages: Map<string, PageRecord> = new Map();
 	last: PageRecord | null = null;
 	swup: Swup;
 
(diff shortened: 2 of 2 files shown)


daun:
Does swup or any official plugin use the `last` property? [I think not](https://github.com/search?q=org%3Aswup+last&type=code). Just wondering because I've never seen that. But anyway, that's another PR 🤠

daun (approved):

src/modules/fetchPage.ts
```diff
@@ -9,7 +9,7 @@ export function fetchPage(this: Swup, data: TransitionOptions): Promise<PageReco
 
 	if (this.cache.exists(url)) {
 		this.triggerEvent('pageRetrievedFromCache');
-		return Promise.resolve(this.cache.getPage(url));
+		return Promise.resolve(this.cache.getPage(url) as PageRecord);
```
gmrchk, line 12: While the casting is fine thanks to the `exists` check above, I think it would be cleaner to retrieve the record and be working with the return value directly, since it can be `null`.

gmrchk:
> Does swup or any official plugin use the `last` property? [I think not](https://github.com/search?q=org%3Aswup+last&type=code). Just wondering because I've never seen that. But anyway, that's another PR 🤠

Likely not, but we need to be careful with these changes. Same goes for switching to `Map`. It could be used by projects or plugins directly, since the public/private distinction is poorly defined right now.

src/modules/fetchPage.ts
hirasso: You mean without the `this.cache.exists` and instead with `this.cache.getPage`?

hirasso:
> Likely not, but we need to be careful with these changes. Same goes for switching to Map. It could be used by projects or plugins directly, since the public/private distinction is poorly defined right now.

@gmrchk Do you also have these reservations if introducing this as a breaking change in swup@4? The advantage of using a `Map` isn't obvious right now, but it will enable more advanced use cases.

src/modules/fetchPage.ts
gmrchk: Yup! At that point, we already know it's `PageRecord` or `null` and can use that as a check instead of `exists` and don't access the cache once more.

daun:
@gmrchk What was the `last` property used for? If it was never documented or advertised, it's probably not in use. It's definitely in use in any plugin, I checked. Not sure about any projects in the wild.

The switch to `Map` itself is definitely not a breaking change in my opinion — all access to the entries happens over the provided methods like `getPage`.

gmrchk:
The `last` property could be used in someone's project, although it's unlikely since it was never mentioned in docs or anywhere. Still, I would consider it a breaking change.

The same goes for switching to `Map`. While we use `getPage` and all of that, the `pages` prop is still public and accessible, so any operation directly on that is basically valid and not wrong.

daun:
In that case, let's unmerge this PR and the switch to `Map` — we should only have changes on `master` that can be published as a minor or patch release. The breaking changes should probably end up in some sort of `next` branch then.

gmrchk:
https://github.com/swup/swup/pull/604

gmrchk:
Agree. I think the general theme of v4 should be (not limited to) making the API stable and cleaned up for outside consumers further.

daun:
@gmrchk I'm stupid. This went into the `next` branch. So it could stay there, I guess?

hirasso:
I disagree - none of us is stupid :)

PR #606, opened 2023-01-24
Turn plugin module tests into unit tests

daun:
- Move everything better suited to unit tests out of cypress
- Should speed tests up a tiny bit

Question

- Do we currently require passing unit tests as part of the CI workflow?


src/__test__/index.test.ts
```diff
@@ -62,3 +75,61 @@ describe('exports', () => {
 		expect(ignoreVisit.mock.calls).toHaveLength(1);
 	});
 });
+
+describe('Plugin module', () => {
```
gmrchk, line 79: Please feel free to move these around into separate files if you wish. This file already starts to grow quite a lot. 
I think these could be in `plugins.test.ts` next to the `plugins.ts` file, for example.

gmrchk (approved):

src/__test__/index.test.ts
daun: Good point, will address this in a separate PR.

PR #607, opened 2023-01-24
Feature: Cache expiration

hirasso:
**Description**

As discussed in #602 , this is a draft for a cache expiration mechanism.

- Each time a page from the cache is requested, the cache is being freed from expired pages before checking for a result
- The time to live for each page can be overwritten with a new option: `getTimeToLive: (page: PageRecord) => number;`
- The default `ttl` is 600 seconds (10 minutes) for all pages
- Returning `0` for a page will never expire it.

The callback option opens the door for very interesting use cases. Primary pages could always stay in the cache by returning `0` from the callback, other pages could get fine-grained control over the `ttl` based on user-defined logic.

**Checks**

### DIFF
--- a/src/modules/Cache.ts
+++ b/src/modules/Cache.ts
@@ -5,6 +5,7 @@ import { PageData } from './getPageData';
 export interface PageRecord extends PageData {
 	url: string;
 	responseURL: string;
+	timestamp: number;
 }
 export class Cache {
 	pages: Map<string, PageRecord> = new Map();
@@ -32,7 +33,12 @@ export class Cache {
 		this.swup.log(`Cache (${this.pages.size})`, this.pages);
 	}
 
+	/**
+	 * Get a page from the cache.
+	 * Will make sure that expired pages won't be returned.
+	 */
 	getPage(url: string): PageRecord | null {
+		this.deleteExpiredPages();
 		return this.pages.get(this.getCacheUrl(url)) || null;
 	}
 
@@ -52,5 +58,35 @@ export class Cache {
 
 	remove(url: string): void {
 		this.pages.delete(this.getCacheUrl(url));
+	}
+
+	/**
+	 * Delete all expired pages
+	 */
+	deleteExpiredPages(): void {
+		for (const [key, page] of this.pages.entries()) {
+			if (this.isExpiredPage(page)) {
+				this.pages.delete(key);
+			}
+		}
+	}
+
+	/**
+	 * Checks if a page should be considered expired.
+	 * Will return true if the page was saved more then 10 minutes ago
+	 */
+	isExpiredPage(page: PageRecord): boolean {
+		const ttl = this.getTimeToLive(page);
+		if (ttl === 0) return false;
+		return new Date(Date.now() - page.timestamp).getSeconds() > ttl;
+	}
+
+	/**
+	 * Get the time to live for a page record.
+	 * Defaults to 600 seconds (10 minutes).
+	 * If this returns 0 the cache for the requested page will never expire
+	 */
+	getTimeToLive(page: PageRecord): number {
+		return 600;
 	}
 }
--- a/src/modules/Cache.ts
+++ b/src/modules/Cache.ts
@@ -76,17 +76,8 @@ export class Cache {
 	 * Will return true if the page was saved more then 10 minutes ago
 	 */
 	isExpiredPage(page: PageRecord): boolean {
-		const ttl = this.getTimeToLive(page);
+		const ttl = this.swup.options.getTimeToLive(page);
 		if (ttl === 0) return false;
 		return new Date(Date.now() - page.timestamp).getSeconds() > ttl;
-	}
-
-	/**
-	 * Get the time to live for a page record.
-	 * Defaults to 600 seconds (10 minutes).
-	 * If this returns 0 the cache for the requested page will never expire
-	 */
-	getTimeToLive(page: PageRecord): number {
-		return 600;
 	}
 }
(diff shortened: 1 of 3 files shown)


src/modules/getPageData.ts
```diff
@@ -19,6 +20,7 @@ export const getPageData = function (this: Swup, request: XMLHttpRequest): PageD
 
 	return {
 		...pageHtmlData,
-		responseURL: request.responseURL || window.location.href
+		responseURL: request.responseURL || window.location.href,
+		timestamp: Date.now()
```
gmrchk, line 24: I think we should provide several timestamps actually, `createdAt`, `lastUpdated` and maybe `lastAccessed`. This way custom logic can be based on whatever suits your use case.

src/modules/Cache.ts
```diff
@@ -53,4 +59,25 @@ export class Cache {
...
+		}
+	}
+
+	/**
+	 * Checks if a page should be considered expired.
+	 * Will return true if the page was saved more then 10 minutes ago
+	 */
+	isExpiredPage(page: PageRecord): boolean {
+		const ttl = this.swup.options.getTimeToLive(page);
+		if (ttl === 0) return false;
```
gmrchk, line 80: Making `0` never delete the cached pages seems counterintuitive. I would expect such a value to always delete the cache (basically act as `cache: false`).

gmrchk (commented):
Looking at the logic, it's quite small, so it wouldn't be a problem when it comes to the bundle size, but it's also very specific, essentially not helping when a custom logic for cache invalidation would be introduced by someone. 
That being said, I'm even more convinced it should be a plugin, possibly with multiple strategies to invalidate the cache, and the only thing that would be added to Swup core is the additional info to work with, like the proposed timestamps, or maybe more. This would also make the custom plugins for such purposes very easy to write.

daun:
Great 🌝

I agree with @gmrchk that this feels more like a useful plugin with good options than part of the core.

Adding `createdAt`, `updatedAt`, `retrievedAt` sounds good, but that can also be done from the plugin itself in a `pageCached` event using the `updateCache()` method. No reason to store that information when not using any invalidation at all. Because it wouldn't necessarily end with these — it might be useful to store `accessCount` or `preloaded` or any other type of metadata we can't think of right now.

hirasso:
Wouldn't it collide with TypeScript if these properties were to be dynamically defined by a plugin?

daun:
@hirasso I'm sure there's a way of extending/inheriting a type and allowing custom properties on top of the defined ones. Not a TS expert, though. It will definitely work if the plugin is plain JS 👽

gmrchk:
I could see the whole system becoming super cumbersome with Cache being provided as an outside entity altogether. For example, to write your own logic for invalidating such cache, you'd need to rely on the plugin to enable the Cache in the first place. Since Swup core wouldn't hold any cache at all, you'd need to also somehow import TS types from such a plugin, to extend Swup with it as well, meaning Swup types would need to be prepared for it (likely accepting some optional generic). It's not impossible, but imagine the `CacheInvalidation` plugin - it would basically have a direct `Cache` plugin dependency, meaning that the plugin would need to be provided to Swup before `CacheInvalidation` (in plugins option array), already leading to constraints that are hard to control in every single place (`CacheInvalidation` needs to check for the existence of some cache, etc.). 

The cleanest way I can see is Swup having a cache by default and using it by default as it does now (with `cache: true`). That cache could hold all kinds of helpful info like the one proposed above and even more if necessary. This can always be extended without breaking changes. The logic manipulating the cache, on the other hand, would be a responsibility of an official plugin providing all kinds of strategies for invalidating, including the one where the cache is kept empty almost at all times (like `cache: false` now). Alternatively, custom logic or a custom plugin can easily manipulate the cache as well, since all the info is in the cache of Swup instance already. 

In simplified language, we would be making the distinction between `information` and `logic` related to cache: 
* Information - always swup core. 
* Logic - always plugin or custom logic or a custom plugin.

daun:
The scenario you mention is indeed cumbersome — having a Cache as well as a CacheInvalidation plugin. That's just overly complex. I think the actual options are different, though:

- A: swup core **with** cache + invalidation plugin
- B: swup core **without** cache + cache plugin incl. invalidation

In both cases, people can decide if the cache/invalidation thing is important enough to them to pull in another package. But no cross-dependencies are required.

Technically, having a cache as a plugin also wouldn't require putting anything on the swup instance itself. A cache plugin could just use the new `fetchPage` method we introduced, return the page object from its own cache map if found, otherwise let swup fetch and afterwards put the page object into its own cache map. Same as the preload plugin now doesn't need to put anything on the swup instance by wrapping the fetchPage method.

hirasso:
@daun since you implemented the overwritable `fetchCache` method: would multiple plugins (cache+preload-plugin) be able to overwrite it without creating conflicts between each other?

daun:
@hirasso That's a really good point; the preload plugin requires the cache to be there. Technically, the setup should work, but only if the preload plugin has `require: { CachePlugin: '>=1.0.0' }` in its definition.

I think the big question is: would we be comfortable introducing dependencies between plugins? If not, we should bury the idea of a general cache plugin and move on 🌝 It was more of a puristic thought experiment, anyway.

gmrchk:
I'm all for the separation when I think about it; the puristic side of me kicks in too. 😄 That's why the whole plugin system was created, after all. The plugin dependency is also kind of prepared for it already with the recent check additions.
But at the same time, I'm skeptical about it having a good user experience, whether it's the plugins depending on each other just a bit too much (leading to cumbersome setup, even when simple) or allowing users to manipulate cache easily programmatically by an included API. 

Plenty of ideas tho, I think nothing left to do than to give some of those a go and see how it goes? I think we can chat about that at the next meeting. 

Regarding this specific proposal, let's close the PR and continue any further comments in the #602.

PR #609, opened 2023-01-26
Allow ignoring visits by original event

daun:
**Description**

- Pass the original `event` that triggered a visit into the `ignoreVisit` callback
- Allow canceling a visit by, e.g.
  - the mouse coordinates of a link click
  - the submitter element of a form submission
- I needed this recently to filter out button-click submissions from enter-key submissions
- The feature was requested by users a few times: having access to the original click event in a callback

**Questions**

- Does this need better types? Using `Event` seems future-proof, but not sure.

**Checks**


src/Swup.ts
```diff
@@ -47,7 +47,7 @@ export type Options = {
 	requestHeaders: Record<string, string>;
 	plugins: Plugin[];
 	skipPopStateHandling: (event: any) => boolean;
-	ignoreVisit: (url: string, { el }: { el?: Element }) => boolean;
+	ignoreVisit: (url: string, { el, event }: { el?: Element; event?: Event }) => boolean;
```
gmrchk, line 50: Do we know what `Event`s could be received here? `MouseEvent` would carry the coordinates, for example, but I'm guessing we don't know what the event would be. 

Just thinking out loud, but maybe that indicates that such functionality shouldn't be handled here. Would it make more sense to introduce this "cancel" functionality into events passed to event handlers? This way we know exactly what the event is (eg. `swup.on('popState', (event /* :PopState */) => {})`). Then we can have something like `event.cancelSwupTransition` on those events.

daun: We don't know the type of event beforehand since it could be `mouseenter` for preloading, `submit` for forms and `click` for normal link visits. Adding a type check works fine to distinguish them: `event instanceof MouseEvent` or `event.type === 'click'` either work.

gmrchk (approved):

PR #612, opened 2023-01-27
Specify exact types for each type of event handler

gmrchk:
**Description**

Since Swup event handler types were so far defined as `Event,` there is no way to use it adequately other than casting it to your event where needed. We also need to find out when the event is defined for the handler and when it isn't.

This adds specific types to the event handlers, which should infer specific types based on the event type you use.

There are no changes to the implementation, only types. Still, we can consider this a breaking change to the API.

```ts
const swup = new Swup();
swup.on('popState', (event) => { // <- always PopStateEvent
	console.log(event);
});
swup.on('clickLink', (event) => { // <- always MouseEvent
	console.log(event);
});
swup.on('animationInDone', (event) => { // <- always undefined
	console.log(event);
});
swup.on('transitionStart', (event) => { // <- PopStateEvent | undefined
	console.log(event);
});

swup.off('popState', (event) => {  // <- PopStateEvent
	console.log(event);
});
swup.off('clickLink', (event) => { // <- MouseEvent
	console.log(event);
});
swup.off('animationInDone', (event) => { // <- undefined
	console.log(event);
});

// these just still work without a handler or event name altogether.
swup.off('clickLink');  
swup.off();
```

**Checks**

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -12,5 +12,6 @@ export default {
 			},
 		],
 	},
-	rootDir: 'src'
+	setupFilesAfterEnv: ['./setup-jest.ts'],
+	modulePathIgnorePatterns: ['<rootDir>/cypress/'],
 };
--- a/setup-jest.ts
+++ b/setup-jest.ts
@@ -1,0 +1,3 @@
+console.log = jest.fn();
+console.warn = jest.fn();
+console.error = jest.fn();
--- a/src/Swup.ts
+++ b/src/Swup.ts
@@ -279,7 +279,7 @@ export default class Swup {
 		this.performPageLoad({ url, customTransition });
 	}
 
-	handleLinkToSamePage(url: string, hash: string, event: MouseEvent) {
+	handleLinkToSamePage(url: string, hash: string, event: delegate.Event<MouseEvent>) {
 		// Emit event and exit early if the url points to the same page without hash
 		if (!hash) {
 			this.triggerEvent('samePage', event);
--- a/src/__test__/index.test.ts
+++ b/src/__test__/index.test.ts
@@ -1,10 +1,6 @@
 import pckg from '../../package.json';
 import Swup, { Options, Plugin } from '../index';
 
-console.log = jest.fn();
-console.warn = jest.fn();
-console.error = jest.fn();
-
 const baseUrl = window.location.origin;
 
 function createPlugin(plugin = {}) {
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,9 +1,10 @@
 import Swup, { Options } from './Swup';
 import { Plugin } from './modules/plugins';
+import { Handler } from './modules/events';
 
 export default Swup;
 
 export * from './helpers';
 export * from './utils';
 
-export type { Options, Plugin };
+export type { Options, Plugin, Handler };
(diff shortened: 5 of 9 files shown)


hirasso:
Nice! 🍰 Wouldn't the `clickLink` event need to receive a `delegate.Event<MouseEvent>` instead of a simple `MouseEvent`? I'm accessing `event.delegateTarget` in my handler wich isn't TS complains about.

hirasso:
> Nice! 🍰 Wouldn't the `clickLink` event need to receive a `delegate.Event<MouseEvent>` instead of a simple `MouseEvent`? I'm accessing `event.delegateTarget` in my handler and TS complains about it.

...same goes for "samePage" and "samePageWithHash". This is what I have to do right now:

```
/**
 * Performs a hard reload on the front page if clicking the logo link and
 * `window.scrollY` is lower than 10
 */
swup.on("samePage", (e: any) => { // need to use "any" here, otherwise delegateTarget will throw
  if (window.scrollY > 10) return;

  const url = new URL(e.delegateTarget.href);
  if (url.pathname !== "/") return;

  window.location.reload();
});
```

hirasso:
With `delegate.Event<MouseEvent>` TS doesn't complain about `e.delegateTarget` anymore, but it does still complain about `e.delegateTarget.href`. Makes sense since `delegateTarget` could be any kind of `Element`. Is there a way in TS to cast `delegateTarget` in the Event? Otherwise, I can do it myself in my handler:
```
onClickLink = (e: delegate.Event<MouseEvent>): void => {
  const target = e.delegateTarget as HTMLAnchorElement; // cast e.delegateTarget
  const pathname = new URL(target.href, document.baseURI).pathname;
  // do something with pathname
}
```

gmrchk:
> With `delegate.Event<MouseEvent>` TS doesn't complain about `e.delegateTarget` anymore, but it does still complain about `e.delegateTarget.href`. Makes sense since `delegateTarget` could be any kind of `Element`. Is there a way in TS to cast `delegateTarget` in the Event? Otherwise, I can do it myself in my handler:
> 
> ```js
> 
> swup.on("clickLink", (e) => {
> 
>   const target = e.delegateTarget as HTMLAnchorElement; // need to cast it to be able to access href without TS complaining
> 
>   const pathname = new URL(target.href, document.baseURI).pathname;
> 
>   // do something with pathname...
> 
> });
> 
> ```

Since we're allowing the users to specify the `linkSelector` option, I don't think we should be adjusting that on our side. The element can be "anything" after all, although we probably rely on it being some kind of link or svg link as well.

hirasso:
Right, makes sense! Also for the forms plugin. Since I'm such a TS newbie I gave it a try, this would have casted `delegateTarget` correctly on the swup side (just posting for my future self, don't want to teach you anything 😅):
```js
clickLink: delegate.Event<MouseEvent> & {
	delegateTarget: HTMLAnchorElement
};
```

hirasso (approved):

gmrchk:
@hirasso I've added an export for the Handler types so it can be used for typing your handlers when defining separately.

```typescript
const popStateHandler: Handler<'popState'> = (event) => { // <- PopStateEvent
	console.log(event);
};
swup.on('popState', popStateHandler);
```

FYI, the delegate is generic and has optional element param too, so this should work better.
 
```typescript
delegate.Event<MouseEvent, HTMLAnchorElement>
```

hirasso:
😳 WTF these generics still look like total hieroglyphs to me 😄 ...Just tested it with my `clickLink` handler, works great!

hirasso (approved):

Issue #601, opened 2023-01-23
Handle very long server response times


hirasso:
**Describe the problem**

I have observed that in rare cases, if the server is momentarily unresponsive, swup gets stuck in loading the page. Manually reloading the page would load the requested page immediately.

**Describe the proposed solution**

Swup should automatically perform the hard reload after a (configurable?) `timeout` in `loadPage`. I would go for a timeout around 10 seconds.

gmrchk:
Retrying the (GET) request after timeout is also a possibility. That could probably be the default behavior adjustable in the handler you suggested. In such a case we should probably be providing a `retry` function to the handler that the user would just call to repeat the request and cancel the previous one, or some other means to easily retry.

daun:
I'd go with a simple timeout and then reload via `window.location`. Easy to maintain, easy to reason about.

For anything else, we lack the information on what failed and why. Retrying might be the worst solution when the server is overloaded, or the best solution when the network is flaky, but we don't really know.

daun:
Is there a way to add a timeout option without a breaking change and major version? I guess not.

hirasso:
There's a bit of an ambiguity here. The current behavior could almost be considered a bug (The site can become unusable, stuck in between transitions, making a user-initialized hard reload necessary). On the other hand, it's of course a change if how swup behaves, so probably still swup@4. I added the label already :)

PR #615, opened 2023-01-30
Request timeout

daun:
(At this point more of a starting point for a discussion than a final solution)

**Description**

- Add new option `timeout` in milliseconds, after which requests are force-loaded via `location.href`
- Defaults to 30 seconds so we can introduce this without breaking change
- Solves #601

**Thoughts/Questions**

- Currently the `timeout` value is applied in `fetchPage`, meaning:
  - The forms plugin will have the timeout applied to its fetches
  - The preload plugin will not have the timeout applied
  - For the preload plugin it makes sense not to have a timeout, though in this case it's just a coincidence that we haven't yet managed to re-use the logic, so ideally it would be shared code
  - For forms, just force-reloading is often the wrong strategy since it would mangle post requests (see next note)
- We have no good way of dealing with POST requests without [creating a form on the fly](https://stackoverflow.com/a/53566845). Should we still force-reload as a GET request here, potentially losing any form data, or do POST requests get a different logic?
- The solution might be to allow opting out of the timeout for POST requests from the `loadPage` method
- Another strategy: a `serverError` callback option that would allow custom handling of timeouts, 500s, 404s, etc.

**Checks**


gmrchk:
Just a quick thought... In the past, the Swup getting stuck was mainly due to incorrectly defined animations. That's not the case anymore, although you can still define your CSS in a way that you'd have a minute running transition by mistake. 
I'm not sure if we should be digging too much into the underlying logic of what to do for POST, GET, plugins, etc. Maybe all the Swup timeout would do is simply be a fail-safe - refresh the page after a set timeout in case the transition has been going on for that time, without checking whether the reason is running fetch, some slow animation, or something else.

daun:
The PR is just concerned with the request to the server itself, so a one-minute animation should still work. I do like keeping it simple, though. We should just know what tradeoffs we're making.

gmrchk:
> The PR is just concerned with the request to the server itself, so a one-minute animation should still work. I do like keeping it simple, though. We should just know what tradeoffs we're making.

I get the concern; there are very few things worse than losing a filled-up form. 😄 How about this: 
* The default is set to `0`, so the current behavior without timeout is used, and there are no breaking changes. Otherwise, I think this can only go into the next major.
* The timeout is applied to all requests regardless of what it is (there are just too many factors for us to be able to decide for everyone).
* `loadPage` method will have a timeout option as well, meaning you can always change the behavior by using Swups API and do the thing programmatically, on click, submit, or whatever.  
* We mention these culprits in the option documentation, where we recommend being careful when submitting forms etc, and adjust the timeout for that if needed.

daun:
@gmrchk Sounds good! I'll make those changes when I'm back at the computer in a week or two. I'd also suggest adding an option to the forms plugin for custom timeout behavior which could help with the post request issue. There's no rush though since we're introducing this as non-breaking.

daun:
Closing in favor of #693

PR #617, opened 2023-02-06
Fix `Referer` request header

### LINKED ISSUE #616: Invalid referer header
**Description of the issue**

Due to history handling logic, `referer` header is _always_ set to the current page being loaded.

As on image below I am requesting `/examples` from different page
image

**How to reproduce the issue**

Issue is clearly visible on swup homepage with javascipt enabled (you can compare with JS disabled, what headers are being sent)

Issue is present in v2-v3.

**Issue cause**

I believe that creating history record should be moved after the page fetch. But I haven't studied the rest of the code to fully understand the reasoning behind current implementation.
https://github.com/swup/swup/blob/343319c986d10506b2ee402c17d8b16cf49a5e1a/src/modules/loadPage.ts#L45-L52

**Severity**

Annoyance, one would expect to behave correctly. But can be worked around with custom request setting on `clickLink` event

hirasso:
Fixes #616

**Description**

Right now, the native `Referer` request header always contains the same URL as the requested page, since `fetchPage` is being executed right **after** the URL was updated through `createHistoryRecord`. This PR attempts to fix this behavior by initializing `fetchPage` right **before** updating the URL.

**Additional information**

Tested this manually, works as expected on my end (correct `Referer` header).

### DIFF
--- a/src/modules/loadPage.ts
+++ b/src/modules/loadPage.ts
@@ -41,15 +41,15 @@ export function performPageLoad(this: Swup, data: PageLoadOptions) {
 	// start/skip animation
 	const animationPromises = this.leavePage({ event, skipTransition });
 
+	// Load page data
+	const fetchPromise = this.fetchPage(data);
+
 	// create history record if this is not a popstate call (with or without anchor)
 	if (!isHistoryVisit) {
 		createHistoryRecord(url + (this.scrollToElement || ''));
 	}
 
 	this.currentPageUrl = getCurrentUrl();
-
-	// Load page data
-	const fetchPromise = this.fetchPage(data);
 
 	// when everything is ready, render the page
 	Promise.all<PageRecord | void>([fetchPromise, ...animationPromises])
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -13,6 +13,9 @@ jobs:
     steps:
       - name: Check out code
         uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
       - name: Run cypress tests
         uses: cypress-io/github-action@v4.2.2
         with:
--- a/cypress/e2e/main.spec.js
+++ b/cypress/e2e/main.spec.js
@@ -11,10 +11,21 @@ describe('Request', function () {
 	});
 
 	it('should send the correct referer', function () {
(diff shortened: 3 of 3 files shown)


src/modules/loadPage.ts
```diff
@@ -41,16 +41,16 @@ export function performPageLoad(this: Swup, data: PageLoadOptions) {
 	// start/skip animation
 	const animationPromises = this.leavePage({ event, skipTransition });
 
+	// Load page data
+	const fetchPromise = this.fetchPage(data);
+
 	// create history record if this is not a popstate call (with or without anchor)
 	if (!isHistoryVisit) {
 		createHistoryRecord(url + (this.scrollToElement || ''));
 	}
```
gmrchk, line 50: It's not related to this PR, and I think the behaviour still works fine since it's a sync code even with he order shuffled in this file (after previous refactoring), but just something to keep in mind... some browsers (like Safari on iPhone) create screenshot of the page which is visible when navigating back. This screenshot is created at the moment of changing the URL with `pushState`, so doing it in the process of transition might lead to page being faded away or similar on the screenshot.

But again, just noting the use case, all good here.

gmrchk (commented):
Nice improvement! ~~Just wondering, what do you use the referer for that you noticed it in a first place? Some analytics were wrong?~~ I didn't notice the linked issue. 👍 

Could you add a unit test for this so we can check what headers are used for fetch after using `loadPage`?

hirasso:
@gmrchk I think releasing this without a unit test would be fine – it's a very simple change. WDYT?

gmrchk:
Ups, totally forgot about this one, soory! Yeah let's go ahead and merge it, but in general coverage is good to increased not decrease. While being simple fix, it's also simple to break in a future by just changing some seemigly irrelevant order.

hirasso:
Totally makes sense 🪚 ...if you want to create the unit test I would certainly study it to learn from it :)

daun:
I've pushed two commits. One adds a test for the correct referer header. I made sure it's failing in the current master branch and passing in this one. The other one adds a test for the other request headers, for good measure and since we've never checked for those.

I had to pin the E2E action node version to v16 since `start-server-and-test` is currently stalling with node 18.

daun (approved):

hirasso:
Wow... Cypress is way over my head. How come everyone always assumes that writing tests is "lame and easy"? Great work @daun !

PR #619, opened 2023-02-16
Update version

hirasso:
I'd like to use #612 in production


hirasso:
Aargh, just realized that `v3.0.4` already contains the new handler types...

PR #638, opened 2023-03-30
Support TypeScript `"moduleResolution": "node16"`

### LINKED ISSUE #637: Support TypeScript `"moduleResolution": "node16"`
**Description of the issue**

If a consumer project's `compilerOptions.moduleResolution` is set to `node16`, swup's types can't be resolved anymore. Here's a screenshot:

Screenshot 2023-03-30 at 10 22 43

And one of the `index.d.ts`:

image

**How to reproduce the issue**

Add this to the `tsconfig.json` of a project that imports "swup":

```json
{
  "compilerOptions": {
    "moduleResolution": "node16",
  }
}
```

Now TypeScript can't resolve swup's  types anymore, since they are all using unqualified imports.

**Proposed Solution**

All types should use qualified exports. So for `index.d.ts`, instead of 

```js
import Swup, { Options } from './Swup';
import { Plugin } from './modules/plugins';
import { Handler } from './modules/events';
export default Swup;
export * from './helpers';
export * from './utils';
export type { Options, Plugin, Handler };
```

it should be:

```js
import Swup, { Options } from './Swup.js';
import { Plugin } from './modules/plugins.js';
import { Handler } from './modules/events.js';
export default Swup;
export * from './helpers.js';
export * from './utils.js';
export type { Options, Plugin, Handler };
```
and so on. I tested this and it ~works with both `moduleResolution: node` as well as `moduleResolution: node16`~. EDIT: Weird... it just worked for a bit with the qualified imports in the d.ts files, but now it doesn't anymore. Not sure what changed... 🤔

**Before creating this issue...**

- [x] Have you provided all helpful information available?
- [x] Have you checked closed issues for similar/related problems?

hirasso:
Fixes #637

**Description**

Switched all imports to fully qualified to support `node16` module resolution. This also automatically generates the right `d.ts` files through microbundle.

**Checks**

- [ ] ~All tests are passing (`npm run test`)~
  jest ist not picking up the new moduleResolution, yet. More googling required...

### DIFF
--- a/src/Swup.ts
+++ b/src/Swup.ts
@@ -1,6 +1,6 @@
 import delegate from 'delegate-it';
 
-import version from './config/version';
+import version from './config/version.js';
 
 import {
 	cleanupAnimationClasses,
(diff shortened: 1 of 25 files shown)


hirasso:
https://news.ycombinator.com/item?id=31343533

hirasso:
So, seemingly this doesn't fix the issue described in #637 , but I still think it would be worth it to switch to the new moduleResolution in swup. It's the future, after all 🧨 ...what do you guys think?

hirasso:
I finally found out what the problem was! The `exports` field in the package.json also needs a `types` field. After adding this:
```json
{
  "exports": {
    ".": {
      "require": "./dist/Swup.cjs",
      "import": "./dist/Swup.module.js",
      "types": "./dist/types/index.d.ts"
    }
  }
}
```
The types are now correctly being resolved both with `moduleResolution` set to `node` as well as `node16`!

hirasso:
I know this PR got quite out of hand... Switching to vitest wasn't actually related to my original problem... Does the team have any objections to using `moduleResoltion: Node16` in combination with vitest from now on?

hirasso:
> I quite like the switch to vitest. While on it, could we move setup-vitest.ts into the ___tests___ folder as well to keep the root folder clean? Fells like that's where it belongs.

Yeah, let's keep the root as clean as possible. There are actually multiple `__tests__` folders right now, so that wouldn't be the right location to put the files. That's why I moved both vitest-related files in their own dedicated folder `./vitest` in c8b2525df86759ff24120c7c03135eaa6ba02525.

> As to whether moduleResolution should be node or node16, I'm pretty ignorant here. Are there any disadvantages or compatibility problems to be expected by adopting node16 in your opinion?

As far as I understand it, using `node16` has only upsides. A big one for me is that it forces us to use fully qualified imports throughout the code base. But I can't be completely sure of course... @daun would you mind checking if the package.json in `version.ts` is still being imported correctly, with tree shaking? Wasn't there some horrible shenanigans going on with this? And why does it need to support WebPack 4?

hirasso:
> Also: is nodenext the same as node16? The official TS docs are pretty light on the topic.

Intuitively I would think that `nodenext` means "always use the bleeding edge node resolution", as opposed to `node16`, which is pinning the version to v16. But that's just me 🤷‍♂️

hirasso:
- [x] Direct import from `/swup/src/index.js` still working in WebPack 5
- [x] Package.json still tree-shoken during `npm run build` (only `version` makes it into the bundle)

daun:
I figure `nodenext` is the way to go for frontend libraries like swup, then.

I'm still maintaining lots of projects in production running on webpack 4 and I see no good reason to upgrade them. If it's not too much hassle, we should keep support, not least of all because it would be a breaking change.

hirasso:
> I'm still maintaining lots of projects in production running on webpack 4 and I see no good reason to upgrade them. If it's not too much hassle, we should keep support, not least of all because it would be a breaking change.

Can you elaborate? Why would this be a breaking change? What do you mean exactly with "this"? You don't import swup's source directly in your webpack 4 projects, do you?

gmrchk (approved):

PR #648, opened 2023-05-07
Make check for valid `transitionend`/ `animationevent` more robust

hirasso:
### Problem

Sometimes the value of `elapsedTime` of a `transitionend`/`animationend` event can be `0`. With the current implementation this will lead to swup throwing an error, because it thinks it's not one of these two events:

```js
const isTransitionOrAnimationEvent = (event: any): event is TransitionEvent | AnimationEvent =>
	!!event.elapsedTime;
// ...
if (!isTransitionOrAnimationEvent(event)) {
    throw new Error('Not a transition or animation event.');
}
```

### Solution

I changed the check to test for the `event.type` instead:

```js
const isTransitionOrAnimationEvent = (event: any): event is TransitionEvent | AnimationEvent =>
	[transitionEndEvent, animationEndEvent].includes(event.type);
```

I also don't think a `new Error` should ever be thrown by swup. So this PR also changes that to a `console.warn` with the`event` as second param for debugging.

**Checks**

### DIFF
--- a/src/modules/getAnimationPromises.ts
+++ b/src/modules/getAnimationPromises.ts
@@ -79,7 +79,7 @@ function getAnimationPromiseForElement(
 			}
 
 			if (!isTransitionOrAnimationEvent(event)) {
-				throw new Error('Not a transition or animation event.');
+				return console.warn('Not a transition or animation event', event);
 			}
 
 			// Skip transitions that happened before we started listening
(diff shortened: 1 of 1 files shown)


hirasso:
...actually, I don't even get how the event could ever *not* be one of `transitionend` or `animationend` 😅 But oh well, better save than sorry right?

daun:
The change looks fine, but I don't understand the problem (yet). 

What's a situation where elapsedTime would be 0? Is that situation something we weren't testing for? We should have seen the exception in the test runs for page visits with immediate transitions, I guess -- unless we catch and bury that somewhere higher up.

hirasso:
Actually, my case is pretty obscure. I'll try to do a little more investigating and report back.

gmrchk (approved):

Issue #652, opened 2023-05-19
`getAnchorElement` sometimes failes


hirasso:
**Description of the issue**

I just noticed that some of the anchor links on the swup docs site are not working. Right now it's broken for all links that contain a question mark, like "Having Trouble?" or "Want to Contribute?".

Turns out this happens due to swup [transforming the hash](https://github.com/swup/swup/blob/master/src/modules/getAnchorElement.ts#L12-L13) before it attempts to match it against an element:

```js
hash = decodeURIComponent(hash);
hash = escapeCssIdentifier(hash);
```

So `having-trouble%3F` for example is being converted to `having-trouble\?` and thus can't match the `id` that actually is `having-trouble%3F`.

@daun, would you mind taking a look? Do you still remember why you introduced these transformations?

daun:
We have tests in place to check this function with UTF anchors. I've researched this a bit back then and that's how to convert between hash and CSS selector encodings.

Could this have gone wrong somewhere else? E.g. when creating and encoding the hashes?

daun:
I think I understand now. IDs aren't supposed to be url-encoded. Only the hash itself is -- since it's part of a URL. If a character can't be a valid part of an ID, we should strip it or slugify it as "-" instead of url-encoding it.

daun:
The solution here is either not url-encoding the id or double-encoding the anchor in the link. Linking to `a%3F` would require an id of `a?`. To make the ID `a3%F` work, the link would need to go to `a#%253F` (since the percent sign itself has to be encoded and decoded itself).

hirasso:
Interesting. When I deactivate swup and use the native browser behavior, it works. Shouldn't swup work just like the browser?

hirasso:
Maybe we could try to match an element before doing the conversion and if one is found return early?

daun:
Let me do some more research on the expected behaviour here — if browsers have a way of detecting url-encoded id attributes and skipping the decoding in those cases, we should definitely implement that as well. 

Simply not converting it is not an option as it would lead to JS exceptions from the CSS selector engine, complaining about invalid CSS attribute characters. That's the reason we've implemented this in the first place: it'll lead to invalid selector syntax. We could wrap a try-catch around it, but that would feel weird.

daun:
Turns out browser behavior isn't consistent here so there's no way of exactly reproducing it.

[...]

PR #653, opened 2023-05-20
Fix anchor selector for scroll target

daun:
Fix for #652 to deal with encoded and unencoded hashes for finding a scroll target.

Tries to mimic WebKit behavior by looking for the raw value of the hash in any `[id]` or `a[name]` attribute, then checking for the url-decoded value.

New tests added.

### DIFF
--- a/cypress/e2e/main.spec.js
+++ b/cypress/e2e/main.spec.js
@@ -434,6 +434,16 @@ describe('Scroll Plugin', function () {
 		cy.shouldHaveElementInViewport('[data-cy=anchor-by-name]');
 	});
 
+	it('should prefer undecoded id attributes', function () {
+		cy.get('[data-cy=link-to-anchor-encoded]').click();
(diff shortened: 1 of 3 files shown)


cypress/e2e/main.spec.js
```diff
@@ -434,6 +434,16 @@ describe('Scroll Plugin', function () {
 		cy.shouldHaveElementInViewport('[data-cy=anchor-by-name]');
 	});
 
+	it('should prefer undecoded id attributes', function () {
+		cy.get('[data-cy=link-to-anchor-encoded]').click();
+		cy.shouldHaveElementInViewport('[data-cy=anchor-encoded]');
+	});
+
+	it('should accept unencoded anchor links', function () {
```
hirasso, line 442: I find the wording a bit confusing here. How about these two?

- "should prefer undecoded anchor links"
- "should accept decoded anchor links"

hirasso (approved):

hirasso:
Working like butter 🦋

PR #664, opened 2023-06-02
Improved event system (work in progress)

daun:
Create a more powerful event system with a few features:

- [x] Allow returning `Promise`s from event handlers to delay execution of internal handlers
- [x] Allow running custom event handlers `before` or `after` the internal handler
- [x] Allow specifying `priority` of custom event handlers
- [x] Allow running a handler `once`
- [x] Allow replacing the internal handler entirely
- [ ] Allow modifying the arguments to the internal handler
- [ ] Pass along a global `context` object with properties of the current lifecycle like `from.url`, `to.url`, `isHistoryVisit`, `shouldAnimate`, `shouldScroll`, etc.
- [x] Add compatibility layer to support current plugins

## Pros

- More flexibility for users and plugins to replace original functionality

## Cons

- Increases bundle size from `4.9k` to `5.6k`
- Added complexity from switch to `async/await`, but manageable with good tests

## Questions

1. Should this be `swup.events` instead of `swup.hooks`?

## API

Return Promise to delay execution:

```js
// Returning a Promise from an event handler will pause execution until resolved
swup.hooks.add('animationOutStart', () => {
    return new Promise((resolve) => setTimeout(resolve, 500))
})

// Not returning a Promise is fine, too -- the handler will execute synchronously
swup.hooks.add('animationOutStart', () => {
    document.body.classListe.add('lorem-ipsum')
})
```

Run handler before original handler:

```js
// These will run before swup's internal handler
swup.hooks.add('animationOutStart', () => {}, { before: true })
swup.hooks.before('animationOutStart', () => {}})
```

Define priority of each handler:

```js
// This will run second, even though registered first
swup.hooks.add('animationOutStart', () => {}, { priority: 1 })

// This will run first, even though registered second
swup.hooks.add('animationOutStart', () => {}, { priority: 2 })
```

Run handler only once:
```js 
// These will only ever run once
swup.hooks.add('animationOutStart', () => {}, { once: true })
swup.hooks.once('animationOutStart', () => {})
```

Replace original internal handler:
```js 
// These will replace the original handler completely
swup.hooks.add('animationOutStart', () => {}, { replace: true })
swup.hooks.replace('animationOutStart', () => {})
```

Compatibility layer:

```js
// This will continue to work
swup.on('animationOutStart', () => {})

// This will continue to work and create a new hook type by calling `swup.hooks.create('customHook')`
swup._handlers.customHook = []
```

A few drive-by improvements:

- ported core over to `async/await`
- added `unmarkSwupElements` helper
- updated test server deps and disable cache
- added ESM test fixture
- bound event handlers without temp variables like `boundPopStateHandler`

## Checks

### DIFF
--- a/cypress/fixtures/esm.html
+++ b/cypress/fixtures/esm.html
@@ -1,0 +1,36 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <title>ESM</title>
+    <link rel="stylesheet" href="/assets/main.css">
+</head>
+<body>
+    <header class="header">
+        <ul>
+            <li><a href="/page-1.html">Page 1</a></li>
+            <li><a href="/page-2.html">Page 2</a></li>
+            <li><a href="/page-3.html">Page 3</a></li>
+        </ul>
+    </header>
+    <main id="swup" class="wrapper transition-default" data-cy="container">
+        <h1>ESM</h1>
+        <p>Lorem ipsum dolor sit amet, consectetur adipisicing elit.
+        Alias assumenda consectetur consequatur enim esse incidunt iusto, magnam
+        maxime molestias nisi perferendis perspiciatis quibusdam quos repellat,
+        vel veniam vero voluptate voluptatem.</p>
+    </main>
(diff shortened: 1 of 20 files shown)


daun:
Closing in favor of #666

Issue #651, opened 2023-05-19
Cache fails for page objects without `reponseURL` property


daun:
**Description of the issue**

The cache fails with an error when the page object has no `responseURL` property. This is fine for most installations but fails when using the custom payload plugin. Surfaced in [#1](https://github.com/swup/custom-payload-plugin/issues/1).

**How to reproduce the issue**

Install custom payload plugin.

**Suggested solution**

Fall back to `page.url` or the current location:

```js
responseURL: page.responseURL || page.url || window.location.href
```

daun:
Solved in #665

PR #665, opened 2023-06-02
Simplify container replacement

daun:
**Description**

Prototype for simplifying how content containers are replaced.

- Containers are parsed from selectors right before they will be replaced
- This will allow modifying the containers on-the-fly, e.g. by the upcoming fragment plugin or by users themselves
- Every container selector will match a single element (querySelector instead of querySelectorAll)
- No need to add `data-swup` attributes on the containers anymore
- We only need the `url` and raw `html` from the xml request object
- Everything else (document title, body class) can be extracted just-in-time
- Remove `responseURL` and switch to more universal `url` everywhere to allow easier switch to `fetch`
- We don't need `getPageData` and `getDataFromHtml` helpers anymore
- Reduces bundle size by 0.25k

Caveats

- This only works if we assume we stop supporting custom payloads
- The body class plugin needs a slight update (see updated test file)

**Checks**


daun:
This would probably fix #651

hirasso:
I like it! Should this maybe be a PR to the `next` branch?

daun:
@hirasso It probably should, but I'm not sure what the current state of the `next` branch is. Is everything on it approved, but only approved for the next major version? In that case, I should rebase on the `next` branch...

hirasso:
As far as I know, the next branch wasn't updated since the last major release.

daun:
Rebased current master onto next branch...

daun:
Crap, we've merged some cache logic improvements since the latest master but never released it...

daun:
Will look into this next week...

hirasso:
Oops! Good find @daun  🔍 
Thank you for looking into it!

daun:
Incorporated the previous cache upgrades, up-to-date again.

PR #666, opened 2023-06-03
Improved event system

daun:
Simplified version of #664 

Create a more powerful event system:

- [x] Switch codebase to `async/await`
- [x] Allow returning `Promise`s from event handlers to delay execution of internal handlers
- [x] Allow running event handlers `before` or `after` the internal handler
- [x] Allow running a handler `once`
- [x] Allow replacing the internal handler entirely
- [x] Add compatibility layer to support current plugins

## Pros

- More flexibility for users and plugins to replace original functionality

## Cons

- Increases bundle size from `4.9k` to `5.6k`

## API

Return Promise to delay execution:

```js
// Returning a Promise from an event handler will pause execution until resolved
swup.events.add('animationOutStart', () => {
    return new Promise((resolve) => setTimeout(resolve, 500))
})

// Not returning a Promise is fine, too -- the handler will execute synchronously
swup.events.add('animationOutStart', () => {
    document.body.classListe.add('lorem-ipsum')
})
```

Run handler before original handler:

```js
// These will run before swup's internal handler
swup.events.add('animationOutStart', () => {}, { before: true })
swup.events.before('animationOutStart', () => {}})
```

Run handler only once:
```js 
// These will only ever run once
swup.events.add('animationOutStart', () => {}, { once: true })
swup.events.once('animationOutStart', () => {})
```

Replace original internal handler:
```js 
// These will replace the original handler completely
swup.events.add('animationOutStart', () => {}, { replace: true })
swup.events.replace('animationOutStart', () => {})
```

Compatibility layer:

```js
// This will continue to work
swup.on('animationOutStart', () => {})

// This will continue to work and create a new hook type by calling `swup.events.create('formSubmitted')`
swup._handlers.formSubmitted = []
```

## Checks


hirasso (approved):
This is too much for me to review. But the tests are there and they are passing, right? 🫣😄 And I love the new functionality! Great work @daun! 

Let's give @Georgy a few days to veto this... otherwise, let's merge and release it in a minor version.

hirasso:
I'm a bit afraid of the box of worms this could open up:
```
swup.events.replace('animationOutStart', () => {})
```
@daun what was the use case you had in mind for that? If it's too much to write it down, but you have one, ignore my comment and tell me next time we talk :)

daun:
@hirasso The `replace` function was one of the main points of this rewrite: to be able to easily replace swup's internal handlers. This is supposed to be a new official way of overwriting swup behavior -- instead of replacing the actual function on the instance, which is an undocumented accidental API.

Currently, the events and handlers themselves aren't created in a way to make the best use of this feature, but we can rethink that if we're happy with going that way in general.

daun:
I think I'll work on a few more elaborated examples to show why we should pick this direction :)

hirasso:
@daun I strongly trust your judgement here. Definitely good to merge from my side. Writing the documentation will probably help a lot with coming up with good use cases.

Issue #668, opened 2023-06-04
Add a way to tell swup to use `replaceState` instead of `pushState` for certain links


hirasso:
**Describe the problem**

When talking with @gmrchk about the fragment plugin, the idea came up that it would be handy to be able to tell swup to use `history.replaceState` instead of `pushState` for selected links. On the demo site, this could make sense for the previous/next nav inside the overlay:

![Screenshot 2023-06-04 at 12 27 20](https://github.com/swup/swup/assets/869813/478ca3f5-f873-46fa-9216-15f44a0bbd83)

**Describe the proposed solution**

My first idea would be a data attribute on links, for example `data-swup-history=replace`. Swup would then use `replaceState` instead of `pushState` when clicking this link.

**Alternatives considered**

- Callback function as an option
- Any other ideas?

**How important is this feature to you?**

- [x] Nice to have

daun:
I would love this as an option on the `loadPage` method. A data attribute could then pre-set that option for links, while we can do it manually if required for e.g. the form plugin.

daun:
I like Turbo's API: There's a dom attribute as well as a param in the `visit` method:

> To specify that following a link should trigger a replace visit, annotate the link with data-turbo-action="replace":
> 
> `<a href="/edit" data-turbo-action="replace">Edit</a>`
> 
> To programmatically visit a location with the replace action, pass the action: "replace" option to Turbo.visit:
> 
> `Turbo.visit("/edit", { action: "replace" })`

daun:
Solved by #670

PR #670, opened 2023-06-04
Allow replacing the current history entry

daun:
Implement a way of telling swup to update the current history entry, instead of creating a new one. Solves #668

API

```js
swup.loadPage({ url: '/about', history: 'replace' })
```

HTML

```html
<a href="/about" data-swup-history="replace">About</a>
```

**Checks**

### DIFF
--- a/cypress/e2e/main.spec.js
+++ b/cypress/e2e/main.spec.js
@@ -345,6 +345,22 @@ describe('History', function () {
 		cy.wrapSwupInstance();
 	});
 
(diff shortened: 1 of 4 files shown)


hirasso:
🎉

hirasso (approved):
Niiice!

hirasso:
Should we first release this or first update the Documentation? It's non-breaking, so my first intuition would be 1. release 2. document

PR #683, opened 2023-06-22
Context object

daun:
Pass along a context object with information about the current page load. This context object is available in all hook handlers, as well as in the swup instance itself. Swup, plugins, and handlers can set flags on the context object to enable/disable certain functionality, e.g. let the fragment plugin disable animations and scrolling on the next visit.

## The context object

The context object holds a few properties, e.g. the current and next url, whether to animate this visit, the custom transition, the history action, info about the triggering element/event/popstate, and scroll info.

```js
{
  from: { url: '/home' },
  to: { url: '/about' },
  animate: true,
  transition: 'fade',
  trigger: {
    el: /* <a> element */,
    event: /* MouseEvent */
  },
  history: {
    popstate: false,
    action: 'push',
  },
  scroll: {
    reset: true,
    target: '#footer'
  }
}
```

## Accessing the context object

The context object is available in all hook handlers:

```js
swup.hooks.on('contentReplaced', (context) => {
  console.log('Scrolling to element', context.scroll.target);
});
```

It's also available from inside swup itself, simplifying things by not having to pass around PopStateEvents and other flags:

```js
export const leavePage = async function (this: Swup) {
  if (!this.context.animate) {
    await this.hooks.trigger('animationSkipped');
    return;
  }
}
```

## What to do with this?

Disable animation by setting a flag:

```js
swup.hooks.before('transitionStart', (context) => {
  context.animate = false;
});
```

Disable scrolling by setting a flag:

```js
swup.hooks.before('transitionStart', (context) => {
  context.scroll.reset = false;
});
```

Set a custom transition:

```js
swup.hooks.before('transitionStart', (context) => {
  context.transition = 'slide';
});
```

Get the original trigger event: click, form submit, etc.

```js
swup.hooks.before('transitionStart', (context) => {
  if (context.trigger.event instanceof MouseEvent) {
   console.log('Original click event', context.trigger.event);
  }
});
```

Check if we're on a history visit:

```js
swup.hooks.before('transitionStart', (context) => {
  if (context.history.popstate) {
    console.log('History visit');
  }
});
```

## Plugins

Fragment plugin: Disable animations and scrolling if a fragment visit is detected:

```js
this.swup.hooks.before('transitionStart', (context) => {
  if (isFragmentVisit) {
    context.animate = false;
    context.scroll.reset = false;
  }
});
```

Route plugin: augment from and to with route property:

```js
this.swup.hooks.before('transitionStart', (context) => {
  context.from.route = this.getMatchingRouteName(context.from.url);
  context.to.route = this.getMatchingRouteName(context.to.url);
});
```

## Checks


hirasso:
OMG this looks beautiful!!

PR #684, opened 2023-06-22
Accept #top as a special fragment identifier

### LINKED ISSUE #674: Accept `#top` as valid fragment for scrolling
**Describe the problem**

The special fragment `#top` will always scroll to the top in browser, whether or not an element with the id `top` exists anywhere in the document. Swup currently does nothing in that case.

Quoting the HTML spec on [scrolling to a fragment](https://html.spec.whatwg.org/#scrolling-to-a-fragment):

> If *decodedFragment* is an ASCII case-insensitive match for the string `top`, then return the top of the document.

Amend the `getAnchorElement` helper to return the `html` element itself (or find a way of returning the number `0`) if no element was found for the fragment `#top`.

- [x] Nice to have
- [ ] Would make my life a lot easier

daun:
**Description**

Accept the special fragment `#top` to scroll to the top of the document, whether or not an element with the id `top` exists anywhere in the document. Quoting the HTML spec on [scrolling to a fragment](https://html.spec.whatwg.org/#scrolling-to-a-fragment):

> If *decodedFragment* is an ASCII case-insensitive match for the string `top`, then return the top of the document.

**Checks**

Closes #674

### DIFF
--- a/cypress/e2e/main.spec.js
+++ b/cypress/e2e/main.spec.js
@@ -493,6 +493,15 @@ describe('Scroll Plugin', function () {
 		cy.shouldHaveElementInViewport('[data-cy=anchor]');
 	});
 
+	it('should scroll to top', function () {
+		cy.get('[data-cy=link-to-self-anchor]').click();
+		cy.shouldHaveElementInViewport('[data-cy=anchor]');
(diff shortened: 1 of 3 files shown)


hirasso (approved):
🐝

PR #689, opened 2023-06-24
Cache pruning

### LINKED ISSUE #602: Garbage Collector for `swup.cache`

daun:
**Description**

Refactor the cache to allow customisable invalidation/pruning.

- Add a `pageCached` hook to allow augmenting cache entries with e.g. a counter, timestamp, etc.
- Simplify the Cache API by staying close to ES Map syntax: has, get, set, delete, clear, size
- Add an `update` method to allow inserting arbitrary data to existing entries
- Add a `prune` method to allow removing entries that match a predicate function
- Removes 4 bytes from the bundle size 🌝

**Pruning**

By default, the new implementation won't add any new data to entries, nor prune it automatically. For consumers that want this, they can use hooks to do that. Or we can create a more sophisticated plugin for cache invalidation.

```js
swup.hooks.on('pageCached', (ctx, { page }) => {

  // When a page is cached, add a timestamp and a TTL
  swup.cache.update(page.url, { ttl: 1000, created: Date.now() });

  // Every ten requests, prune the cache by checking the timestamp
  if (Math.random() < 0.1) {
    swup.cache.prune((url, page) => page.ttl < Date.now() - page.created);
  }

});
```

**Other drive-by improvements**

- Rename `pageLoadedFromCache` for simplicity
- Remove `cache.getCurrentPage`: we now have the page object in the `replaceContent` hook

**Checks**

**Additional information**

Closes #602

### DIFF
--- a/cypress/support/commands.js
+++ b/cypress/support/commands.js
@@ -53,8 +53,8 @@ Cypress.Commands.add('shouldHaveCacheEntries', (urls) => {
 Cypress.Commands.add('shouldHaveCacheEntry', (url) => {
 	cy.window().should((window) => {
 		const { cache } = window._swup;
-		const exists = cache.exists(url);
-		const page = cache.getPage(url);
+		const exists = cache.has(url);
+		const page = cache.get(url);
 		expect(url).to.be.a('string');
 		expect(exists).to.be.true;
 		expect(page).not.to.be.undefined;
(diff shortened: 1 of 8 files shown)


hirasso (approved):
🍰

PR #691, opened 2023-06-26
Feat: Normalize and compare URLs

### LINKED ISSUE #690: URLs should be normalized before comparing them
**Describe the problem**

Sometimes swup needs to compare two URLs (`pathname` + `search`). Right now, this can lead to false negatives, since the URLs could look different, but actually point to the same resource.

URLs with or without a trailing slash in the `pathname` are being considered different: 

```js 
"/my-path" === "/my-path/"`
```

```js
"/my-path?foo=bar" === "/my-path/?foo=bar"`
```

Also, the order of the searchParams matters when it shouldn't:

```js
"/my-path?foo=bar&baz=bat" === "/my-path/?baz=bat&foo=bar"`
```

The current behavior could lead to cache misses and also poses problems for the fragment plugin.

I would propose we normalize the URLs every time they are being get or set.

1. Apply the normalization in the `fromElement`/`fromUrl` methods of the `Location` class.
2. Use the `Location` class in `getCurrentUrl()`, as well

**Alternatives considered**

To be completely sure that two URLs will be normalized before being compared, we could also move the normalization logic in a comparison function, like `isEqualUrl`. Maybe a combination of both, normalization on initialization and YIT normalization is the most robust solution, to help prevent future regressions.

- [x] Nice to have
- [x] Would make my life a lot easier

hirasso:
Closes #690 

**Description**

Normalize URLs to make them comparable: 

- Remove possible trailing slash from `pathname`
- Sort `searchParams`

**Tasks**

- [x] Normalize URLs in the `Location` helper
- [x] Normalize `getCurrentUrl()`
- [ ] Add method to check two locations for equality

**Checks**

- [x] The PR is submitted to the `next` branch, since normalized URLs could break external expectations

### DIFF
--- a/cypress/e2e/main.spec.js
+++ b/cypress/e2e/main.spec.js
@@ -336,7 +336,7 @@ describe('Link selector', function () {
 	});
 });
 
-describe.only('Resolve URLs', function () {
+describe('Resolve and normalize URLs', function () {
 	beforeEach(() => {
 		cy.visit('/page-1.html');
 		cy.wrapSwupInstance();
--- a/cypress/e2e/main.spec.js
+++ b/cypress/e2e/main.spec.js
@@ -21,11 +21,13 @@ describe('Request', function () {
 		const expected = this.swup.options.requestHeaders;
 		cy.intercept('GET', '/page-3.html').as('request');
 		cy.triggerClickOnLink('/page-3.html');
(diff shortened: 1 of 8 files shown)


hirasso:
Great that you have learned all this the hard way, already! I'll close this PR then and create a simpler helper instead.

PR #692, opened 2023-06-26
Native fetch

daun:
**Description**

Next try of #569 after the rewrite to hooks.

- [x] Switch to native `fetch` API
- [x] Create a new hook `fetchPage` that plugins can hook before to return a fetch response
- [x] Create a new hook `loadPage` that plugins can hook before to return a page object
- [x] Create a new error type `FetchError` to identify fetch errors in loadPage
- [x] Test if server errors (500) are propagated properly as before
- [ ] Test this with plugins (forms and preload plugins specifically)
    - [ ] Make sure none of the plugins depend on the previously returned `xmlHttpRequest` directly
- [ ]  ~~Allow aborting requests~~ — will do later as part of `timeout` feature

**Plugin hooks**

This is how the preload plugin will now inject preloaded pages:

```js
this.swup.hooks.before('loadPage', (context, args) => {
  args.page = this.preloadPromises.get(args.url);
});
```

**Checks**

**Closes**

#566


hirasso:
Aborting requests came up to gracefully handle hung up requests

daun:
@hirasso Was that in relation to the planned `timeout` feature? I also remember us trying to make the preload promises abortable, but it's all been a while and my memory is hazy 😶‍🌫️

hirasso:
Exactly, some kind of timeout was the plan IIRC :)

cypress/e2e/main.spec.js
```diff
@@ -27,6 +27,57 @@ describe('Request', function () {
...
+	it('should force-load on network error', function () {
+		cy.intercept('/error-network.html', { times: 1 }, { forceNetworkError: true });
+		cy.shouldHaveReloadedAfterAction(() => {
+			this.swup.loadPage('/error-network.html');
+		});
+		cy.shouldBeAtPage('/error-network.html');
+	});
+});
+
+describe.only('Fetch', function () {
```
hirasso, line 48: `.only` loves to sneak into PRs :)

src/helpers/Location.ts
```diff
@@ -5,30 +5,33 @@
...
 	/**
 	 * Instantiate a Location from an element's href attribute
-	 * @param {Element} el
+	 * @param el
 	 * @return new Location instance
 	 */
-	static fromElement(el: HTMLAnchorElement): Location {
+	static fromElement(el: Element): Location {
 		const href = el.getAttribute('href') || el.getAttribute('xlink:href');
 		return new Location(href!);
```
hirasso, line 26: Maybe a drive-by improvement could be to make sure that the href is always a string here:
```js
const href = el.getAttribute('href') || el.getAttribute('xlink:href') || '';
return new Location(href);
```

src/Swup.ts
```diff
@@ -332,21 +332,21 @@ export default class Swup {
...
-			popstate: true
-		});
 
 		// Does this even do anything?
 		// if (!hash) {
 		// 	event.preventDefault();
 		// }
 
 		this.hooks.triggerSync('popState', { event }, () => {
+			this.context = this.createContext({
```
hirasso, line 342: Could you briefly explain why the context must be created inside the handler instead of before? Just for me to get a better feel for the concept.

src/modules/Hooks.ts
```diff
@@ -16,17 +16,18 @@ export interface HookDefinitions {
 	clickLink: { event: DelegateEvent<MouseEvent> };
 	disabled: undefined;
 	enabled: undefined;
+	fetchPage: { url: string; options: FetchOptions; response?: Response | Promise<Response> };
+	loadPage: { url: string; options: FetchOptions; page?: PageData | Promise<PageData> };
```
hirasso, line 20: Are you still planning to refactor `loadPage`, `performPageLoad` et al?

src/modules/fetchPage.ts
```diff
@@ -1,54 +1,81 @@
...
+		this.name = 'FetchError';
+		this.url = details.url;
+		this.status = details.status;
+	}
+}
+
+/**
+ * Fetch a page from the server, return it and cache it.
+ */
+
```
hirasso, line 29: an empty line

src/modules/fetchPage.ts
```diff
@@ -1,54 +1,81 @@
...
-			if (requestURL === url) {
-				this.cache.set(url, page);
-			}
-
-			this.hooks.trigger('pageLoaded', { page });
-			resolve(page);
-		});
-	});
+	// Resolve real url after potential redirect
+	const { url: finalUrl } = new Location(responseUrl);
```
hirasso, line 68: I keep being irritated by the naming of `Location`. I always think it's the `window.location` or something, even if that of course doesn't make sense. Anyways, a bit late to rename it now, isn't it? 🐌 `SwupLocation` or `SwupURL` would have been more clear IMHO.

hirasso (commented):
Nice work! It's way to much to test without actually using it. Looking forward to a first release candidate! 🥳

cypress/e2e/main.spec.js
daun: 🫠

src/Swup.ts
daun: To avoid creating context in case the `popState` hook is replaced and the visit is never performed. But now that you mention it, I think it's better to revert this and make sure we're creating the context correctly, even if somebody replaces the hook.

src/modules/Hooks.ts
daun: Yes, I'd like to do but in another PR. This one is getting a bit too large.

src/modules/fetchPage.ts
daun: I think `Location` is perfect. While `window.location` is the very specific `Location` the window is currently at, our `Location`s are free-floating ones that aren't attached to anything at the moment, just pointing at things.

For what it's worth, turbo calls this a `Locatable` which is slightly more weird but kind of the same thing.

PR #693, opened 2023-06-28
Request timeout

### LINKED ISSUE #601: Handle very long server response times

daun:
**Description**

- Add a new option `timeout` in milliseconds, after which requests are force-loaded via `location.href`
- Defaults to no timeout: no breaking change
- `fetchPage` and `navigate` can override this based on the use case (preloading, forms, ...)
- Create a new hook `fetch:timeout`
- Adds 0.2K to the bundle — debatable if this is worth it

**Previous discussion**

> - The timeout is applied to all requests regardless of what it is (there are just too many factors for us to be able to decide for everyone).
> - The `navigate` method will have a timeout option as well, meaning you can always change the behavior by using Swups API and do the thing programmatically, on click, submit, or whatever.
> - We mention these culprits in the option documentation, where we recommend being careful when submitting forms etc, and adjust the timeout for that if needed.

**Checks**

**Further information**

Closes #601


hirasso:
I'm ok with leaving this out for now. Enough on the plate for v4

daun:
Sounds good.

daun:
@hirasso Rebased and ready for review :)

hirasso (approved):

PR #694, opened 2023-06-28
feat: re-export path-to-regexp `match`

### LINKED ISSUE #662: Bundle path-to-regexp helper
**Describe the problem**

Three official plugins are now making use of the [path-to-regexp](https://www.npmjs.com/package/path-to-regexp) library for matching URL patterns (including the work-in-progress fragment plugin). This could lead to duplicate versions being bundled by consumers of these plugins.

Require `path-to-regexp` from swup core, create a helper around it with a stable API, and offer it as a named export that plugins can consume. Example usage:

```js
import { matchPath } from 'swup';

export default class SomePlugin {
  constuctor({ routes }) {
    this.patterns = routes.map(route => matchPath(route));
  }
}
```

This is identical to how we've dealt with `delegate-it` for event delegation. It won't be bundled into the output, but will be available as a named export.

**Alternatives considered**

- Leave this to the plugins themselves, risking duplicate versions

- [x] Nice to have
- [ ] Would make my life a lot easier

hirasso:
Closes #662 

**Description**

Exports a thin wrapper around [path-to-regexp/match](https://github.com/pillarjs/path-to-regexp/tree/master#match)

That function can then be used by plugins or even swup end-users like this:

```js
import Swup, { matchPath } from 'swup';
const urlMatch = matchPath('/users/:user');
const isMatch = urlMatch('/users/bob');
if (isMatch) // do something...
```

**Checks**

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,7 @@
         "@cypress/code-coverage": "^3.10.0",
         "@swup/browserslist-config": "^1.0.0",
         "@swup/prettier-config": "^1.0.0",
-        "cypress": "^12.3.0",
+        "cypress": "^12.16.0",
         "http-server": "^14.1.1",
         "husky": "^8.0.0",
         "istanbul-lib-coverage": "^3.2.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,8 @@
       "license": "MIT",
       "dependencies": {
         "delegate-it": "^6.0.0",
-        "opencollective-postinstall": "^2.0.2"
+        "opencollective-postinstall": "^2.0.2",
+        "path-to-regexp": "^6.2.1"
       },
       "devDependencies": {
         "@babel/preset-typescript": "^7.18.6",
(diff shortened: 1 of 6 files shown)


daun (approved):

daun:
Looks good. I might have broken the tests with my YOLO push to `next`, exporting the hook types?

daun:
Right, no, you're saying they're passing locally. Same here. No idea.

daun:
I'd say no. It's probably easier to debug in this PR — I wouldn't know how to trigger the tests from the `next` branch out of a PR context.

hirasso:
Looks like we ran out of free tests for this month:

https://github.com/swup/swup/actions/runs/5409060447/jobs/9828760340?pr=694#step:4:140

Screenshot 2023-06-29 at 08 22 17

hirasso:
> Looks like we ran out of free tests for this month:
> 
> https://github.com/swup/swup/actions/runs/5409060447/jobs/9828760340?pr=694#step:4:140
> 
> Screenshot 2023-06-29 at 08 22 17

Seems like that's not it. I created an empty test PR #696 that also [announces the free plan limit](https://github.com/swup/swup/actions/runs/5410863418/jobs/9832883470?pr=696#step:4:138), but passes otherwise. So the failing tests must be related to this specific PR 🤔

hirasso:
Found the problem: The tests are passing when I disable the named export for `matchPath`:

```js
// import { matchPath } from './helpers/matchPath.js';
// export { matchPath };
```
I guess this might be related to the compiler warning that we have been getting for a while:
> Entry module "src/index.ts" is using named and default exports together. Consumers of your bundle will have to use `Swup["default"]` to access the default export, which may not be what you want. Use `output.exports: "named"` to disable this warning

It seems like the UMD bundle isn't exporting `Swup` directly anymore as soon as we are adding the first named export. That makes the tests fail: 

> Swup is not a constructor

Weird thing is that the tests are passing just fine locally 🤷‍♂️

I would LOVE to drop support for the UMD bundle all together – what are the reasons we are still keeping it around?

hirasso:
OMG - that last bit was it! Exporting the helper from index instead of swup solved it. Now I also understand the architecture better! 

I'll close this PR and open another one without all the test commits.

PR #703, opened 2023-07-03
Allow setting transition class targets

daun:
Allow `html` (default) or the containers themselves to receive `is-animating`, `is-leaving`, etc.

This lead to a new option `animationScope: 'html' | 'containers'` 🛸

### DIFF
--- a/src/Swup.ts
+++ b/src/Swup.ts
@@ -41,7 +41,7 @@ export type Options = {
 	skipPopStateHandling: (event: any) => boolean;
 	ignoreVisit: (url: string, { el, event }: { el?: Element; event?: Event }) => boolean;
 	resolveUrl: (url: string) => string;
-	transitionRoot: boolean;
+	transitionRoot: 'html' | 'containers';
 };
 
 export default class Swup {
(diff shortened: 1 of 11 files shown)


src/modules/Classes.ts
```diff
@@ -0,0 +1,39 @@
...
+		this.swup = swup;
+	}
+
+	get selectors(): string[] {
+		const { targets } = this.swup.context.transition;
+		return targets || ['html'];
+	}
+
+	get targets(): Element[] {
+		return Array.from(document.querySelectorAll(this.selectors.join(',')));
```
hirasso, line 18: We should rather iterate the array of containers here and execute `querySelector` for each of them. Otherwise, weird things could happen

src/Swup.ts
```diff
@@ -45,6 +41,7 @@ export type Options = {
 	skipPopStateHandling: (event: any) => boolean;
 	ignoreVisit: (url: string, { el, event }: { el?: Element; event?: Event }) => boolean;
 	resolveUrl: (url: string) => string;
+	transitionRoot: boolean;
```
hirasso, line 44: I think instead of a boolean I would prefer a more descriptive `"html" | "containers"` here

hirasso (commented):
Not a full review, only two thoughts

daun:
> Not a full review, only two thoughts

Happy to implement both once you confirm this is solving our use case with the fragments.

hirasso:
Absolutely 👍

src/modules/Classes.ts
hirasso: done

src/Swup.ts
hirasso: Renamed to `animationScope: 'html' | 'containers'`

PR #709, opened 2023-07-06
Pre-release action

daun:
- Publish a prerelease to the npm `next` channel
- Runs on every push to a special `prerelease` branch


daun:
Looks like it doesn't do anything :(

hirasso:
☹️😄

PR #714, opened 2023-07-08
Allow awaiting page load before starting animations

daun:
**Description**

- Add a new flag to the context object: `transition.wait`
- Determines whether to wait for the next page to finish loading before starting the out-animation
- Defaults to `false` and the current behavior: start page load and out-animation at the same time
- Use case: in the upcoming parallel plugin, we need to wait for the page to load before starting both in- and out-animation at the same time

**Checks**

### DIFF
--- a/src/modules/Context.ts
+++ b/src/modules/Context.ts
@@ -17,6 +17,7 @@ export interface PageContext {
 
 export interface TransitionContext {
 	animate: boolean;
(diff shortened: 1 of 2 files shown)


hirasso (approved):
We should really rename `context.transition` to `context.animation` :)

daun:
💯

PR #716, opened 2023-07-09
Rename animation context

### LINKED ISSUE #715: Rename `context.transition` to `context.animation`
As mentioned in https://github.com/swup/swup/pull/714#pullrequestreview-1520705444

daun:
**Description**

- Rename `context.transition` to `context.animation` to better communicate intent
- This is a bit confusing as we use transition and animation almost completely interchangeably, e.g. in `data-swup-transition`, so it might not make things much clearer anyway?

**Checks**

**Additional information**

Closes #715

### DIFF
--- a/cypress/e2e/main.spec.js
+++ b/cypress/e2e/main.spec.js
@@ -660,7 +660,7 @@ describe('Context', function () {
 
 	it('should allow disabling animations', function () {
 		this.swup.hooks.before('transitionStart', (context) => {
-			context.transition.animate = false;
+			context.animation.animate = false;
 		});
 		cy.transitionWithExpectedDuration(0, '/page-2.html');
 	});
--- a/src/index.ts
+++ b/src/index.ts
@@ -4,7 +4,7 @@ import type {
 	Context,
 	FromContext,
 	ToContext,
-	TransitionContext,
+	AnimationContext,
 	ScrollContext,
 	HistoryContext
 } from './modules/Context.js';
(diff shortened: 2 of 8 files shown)


daun:
@hirasso Do you think this is worth it? Considering that it's still called `data-swup-transition` and custom transition name? I feel like the confusion won't go away from this.

hirasso:
I like this a lot! In the case of `data-swup-transition` we could think about introducing an alias `data-swup-animation` and document that instead? Or we introduce a breaking change to keep the bundle the same size. What do you think?

daun:
I'll have to go through the docs once more to get my head around the naming. Just realized it's also called `animationClass: transition-*` 🤯

hirasso:
> I'll have to go through the docs once more to get my head around the naming. Just realized it's also called animationClass: transition-* 🤯

Yeah but that actually makes sense. In the example, it's applying a CSS `transition`. So the selector reflects that. If one would want to use css keyframe `animations` instead, they could use `animationSelector: animation-*` or whatever.

hirasso (approved):

daun:
I've prepared a [PR on the docs](https://github.com/swup/docs/pull/124) to make the naming around this more strict.

daun:
I've pushed some changes to be much more strict about naming things in the code as well. Switched to `data-swup-animation` for uniformity. I'd love to get as many breaking changes into swup 4 now that we've given them some thought.

PR #718, opened 2023-07-11
Refactors

hirasso:
**Description**

A few minor changes to make the code more explicit. Also, unit tests for `replaceContent`

**Checks**

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
     "animation"
   ],
   "dependencies": {
+    "@types/jsdom": "^21.1.1",
     "delegate-it": "^6.0.0",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -59,6 +59,7 @@
     "animation"
   ],
   "dependencies": {
+    "@types/jsdom": "^21.1.1",
```
daun, line 62: I think this should be a devDependency so frontend users don't end up downloading it.

daun:
Nice! Proper tests for single functions 🏓

daun:
Ready to merge if we can move the jsdom types to dev dependencies.

PR #722, opened 2023-07-12
Check direction of travel on history visits

daun:
**Description**

- Calculate the direction of travel for history visits: `forward` or `backward` or undefined
- Available in `context.history.direction`
- Is this useful? Is this worth the 70 bytes? I'd say yes
- Alternative API: numerical delta in `context.history.delta`: `-4`, `0`, `1`, and so on

**Checks**

### DIFF
--- a/src/modules/visit.ts
+++ b/src/modules/visit.ts
@@ -4,6 +4,7 @@ import { FetchOptions } from './fetchPage.js';
 import { ContextInitOptions } from './Context.js';
 
 export type HistoryAction = 'push' | 'replace';
+export type HistoryDirection = 'forward' | 'backward';
 
 export type PageLoadOptions = {
 	/** Whether this visit is animated. Default: `true` */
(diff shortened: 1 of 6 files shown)


src/modules/visit.ts
```diff
@@ -4,6 +4,7 @@ import { FetchOptions } from './fetchPage.js';
 import { ContextInitOptions } from './Context.js';
 
 export type HistoryAction = 'push' | 'replace';
+export type HistoryDirection = 'forward' | 'backward';
```
hirasso, line 7: I always use "forwards"/"backwards" 🤷‍♂️😄

hirasso (approved):
I see there are tests! Lets ship it and see how it works 🫶

src/modules/visit.ts
daun: I've loosely based this on the history api which calls it `forward`. But then it again, it also calls it `back` 😵‍💫

PR #724, opened 2023-07-19
Rename `leavePage` and `enterPage`

hirasso:
**Description**

The names of `leavePage` and `enterPage` are hard for me to understand without looking at the implementation. I'd like to rename them so they describe as clearly as possible what they do:

- `leavePage` ➡️ `animatePageOut`
- `enterPage` ➡️ `animatePageIn`

**Checks**

- ~The documentation was updated as required~

### DIFF
--- a/src/Swup.ts
+++ b/src/Swup.ts
@@ -13,9 +13,9 @@ import { getAnchorElement } from './modules/getAnchorElement.js';
 import { awaitAnimations } from './modules/awaitAnimations.js';
 import { visit, performVisit, HistoryAction } from './modules/visit.js';
 import { fetchPage } from './modules/fetchPage.js';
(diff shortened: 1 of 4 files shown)


daun:
I like it! I've done some refactoring lately, so now they actually only do the animating which fits perfectly with the proposed names.

daun:
I'd suggest renaming the other two involved methods as well for clarity. What do you think about these?

- `renderPage` → `updatePageContent`
- `replaceContent` → `replaceContainers`

Then we'd only need to move the `document.title` part in `replaceContainers` somewhere else since it doesn't make sense from the naming anymore.

daun:
No, actually, these two are a bit more complicated as they do more than update content or containers. Let's leave them untouched for now!

PR #725, opened 2023-07-19
Force reflow

daun:
**Description**

- Switch to [forcing reflow](https://stackoverflow.com/a/21665117/3759615) after toggling animation classes, instead of awaiting the next tick
- Continues immediately and avoids some weird race conditions
- More noticable when used with parallel animations
- Trick is battle-tested in Vue's transitions ([source](https://github.com/vuejs/core/blob/a3dddd62059f4a7a762046c1e7f01485b96e737d/packages/runtime-dom/src/components/Transition.ts#L240))

**Checks**


daun:
Needs some more testing.

PR #726, opened 2023-07-19
Allow skipping animation hook

daun:
**Description**

- Add `skip` arg to `animation:await` hook
- Setting it to false in a `before` hook allows skipping a specific in/out animation

```js
swup.hooks.before('animation:await', (context, args) => {
  if (args.direction === 'out') {
    args.skip = true;
  }
});
```

**Checks**

**Additional information**


hirasso:
Seems to work nicely!! Haven't tested it together with the Parallel Plugin, yet, but if you are confident that this works, I'm good to go.

hirasso (approved):

PR #748, opened 2023-07-28
Prevent accidental cache mutation

hirasso:
**Description**

Returns shallow copies instead of references for `cache.get()` and `cache.all`, to prevent unintentional cache mutation and decrease the API surface. Right now, there are two ways to mutate the cache:

```typescript

function updateCache(url: string, html: string): void {
  const page = swup.cache.get(url);
  if (!page) return;
  
  // Works, official way. Will even create the cache entry if it doesn't exist
  swup.cache.update(url, { html } as CacheData);

  // Also works, could happen unintentionally
  page.html = html;
}

```

**Drive-By**

`cache.update` won't need type casting anymore if only updating the html.

Before:

```js
swup.cache.update(url, { html } as CacheData);
swup.cache.update(url, { html, foo: "bar" } as CacheData);
```

After:

```js
swup.cache.update(url, { html });
swup.cache.update(url, { html, foo: "bar" });
```

**Checks**

### DIFF
--- a/src/modules/Cache.ts
+++ b/src/modules/Cache.ts
@@ -25,7 +25,11 @@ export class Cache {
 
 	/** All cached pages. */
 	get all() {
-		return structuredClone(this.pages);
+		const copy = new Map();
+		this.pages.forEach((page, key) => {
+			copy.set(key, { ...page });
+		});
+		return copy;
 	}
 
 	/** Check if the given URL has been cached. */
--- a/src/modules/Cache.ts
+++ b/src/modules/Cache.ts
@@ -53,9 +53,9 @@ export class Cache {
 	}
 
 	/** Update a cache record, overwriting or adding custom data. */
-	update(url: string, payload: Omit<CacheData, 'url'>) {
+	update(url: string, payload: Record<string, any>) {
 		url = this.resolve(url);
-		const page = { ...this.get(url), ...payload, url };
+		const page = { ...this.get(url), ...payload, url } as CacheData;
 		this.pages.set(url, page);
 	}
 
--- a/src/modules/Cache.ts
+++ b/src/modules/Cache.ts
@@ -33,11 +33,11 @@ export class Cache {
 		return this.pages.has(this.resolve(url));
 	}
 
-	/** Return the cached page object if cached. */
+	/** Return a shallow copy of the cached page object if available. */
 	get(url: string): CacheData | undefined {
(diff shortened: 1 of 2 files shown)


daun:
`structuredClone` looks like it has pretty bad browser supports. Chrome 98, Safari 15.4. Can we use a simple destructuring `{...page}` instead?

hirasso:
Yes we can. Maybe a bit overkill to think about nested custom objects in the cache anyways 😄

hirasso:
Hmm, spreading the result in `cache.all` won't have any effect at all, since it's a Map of objects...

hirasso:
Actually, maybe there is a way: https://stackoverflow.com/a/36392307/586823

daun:
Haha, that sounds almost too good to be true. `new Map(map)`. Great find!

hirasso:
It actually is too good to be true: https://stackoverflow.com/questions/8206988/how-to-copy-a-map-into-another-map/36392307#comment103303505_36392307

daun:
I'm fine with doing shallow copies -- not least for performance. If someone hooks into `cache:set` and reads all entries every time, we might be creating deep clones of hundreds of pages in the middle of a JS animation.

hirasso:
How should the `payload` argument be typed for the best ergonomics? 

**Option 1 (current implementation)**

Needs type casting _always_:

```js
update(url: string, payload: CacheData) {
	url = this.resolve(url);
	const page = { ...this.get(url), ...payload, url };
	this.pages.set(url, page);
}
```
Usage:
```js
swup.cache.update(url, { html } as CacheData);
```

**Option 2**

Allow to update at least the `html` without type casting from the user:

```js
update(url: string, payload: Omit<CacheData, 'url'>) {
	url = this.resolve(url);
	const page = { ...this.get(url), ...payload, url };
	this.pages.set(url, page);
}
```
Usage:
```js
swup.cache.update(url, { html });
swup.cache.update(url, { html, foo: "bar" } as CacheData);
```
**Option 3**

Allow any type of object and cast it in the function:

```js
update(url: string, payload: Record<string, any>) {
	url = this.resolve(url);
	const page = { ...this.get(url), ...payload, url } as CacheData;
	this.pages.set(url, page);
}
```
Usage:
```js
swup.cache.update(url, { html });
swup.cache.update(url, { html, foo: "bar" });
```

daun:
Option 3 looks great!

daun (approved):

PR #753, opened 2023-07-29
Update changelog

daun:


CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
 
 <!-- ## [Unreleased] -->
```
hirasso, line 3: Why do you include this line in all the changelogs?

hirasso (approved):

CHANGELOG.md
daun: We could (theoretically) uncomment the line and collect all unreleased changes below it. Then, when releasing, we can just move the changes down to an actual new release. I wish I'd remember to do that 🐤

Issue #735, opened 2023-07-21
Allow navigation to current location


opencoca:
When users are handling forms, it would be elegant to allow both POSTing the same URL and clearing of form data by just visiting a link to the form. 

**Alternatives considered**

Developers can implement both form-clearing js code and have forms submitted to slightly different URLs than the forms.  IE using a ?dummy_key in the URL and/or a #optional_anchor allows for this functionality without any changes but is a little kludgy. 

**How important is this feature to me?**

- [x] Nice to have
- [x] Would make using swup lot easier for people who are getting started

daun:
Welcome! This makes sense to support, I think. Just to understand your use case: how would you manage 'visiting a link to the form' and have it do a POST request? I assume by using `swup.loadPage()` as the same-page limitation is only in place when clicking links.

daun:
I just had a look at the complexity of this and it's not very likely we'll implement this for the upcoming v4. The concept of ignoring visits to the current URL on link clicks is sort of baked into swup and not very easy to make configurable. This will likely end up in the next major release, though.

In the meantime: this should work as well, although you might end up seeing a scroll to the top, followed by a navigation:

```js
swup.on("samePage", () => {
  swup.loadPage({ url: window.location.href })
})
```

daun:
We've given this a try and it turns out it's very doable! In the upcoming swup v4, you will be able to make swup following links to the current page by replacing the internal handler of the `link:self` hook:

```js
// Make swup follow links to current page
swup.hooks.replace('link:self', (visit) => swup.navigate(visit.to.url))

// Force browser to load links to current page
swup.hooks.replace('link:self', (visit) => window.location.href = visit.to.url)
```

In swup v3, the following should work, although you might end up seeing a scroll to the top, followed by a navigation:

```js
swup.on("samePage", () => swup.loadPage({ url: window.location.href }))
```

I'll close this as implemented.

opencoca:
Thank you for this!

PR #756, opened 2023-08-08
Configure behavior on links to current page

daun:
**Description**

- Add a new option to configure how swup handles links to the current page: `linkToSelf`
- Scroll to the top (current behavior and new default), navigate via swup, or let the browser reload
- Added tests for all three options
- Thoughts on the name of the option? Tried to be consistent with the `link:self` hook

```ts
export type Options = {
  /** How swup handles links to the same page. Default: `scroll` */
  linkToSelf: 'scroll' | 'navigate' | 'reload';
};
```

**Checks**

### DIFF
--- a/cypress/e2e/main.spec.js
+++ b/cypress/e2e/main.spec.js
@@ -362,6 +362,21 @@ describe('Link resolution', function () {
 			expect(navigated).to.be.false;
 		});
 	});
+
+	it('should navigate to same page if configured via linkToSelf option', function () {
+		let navigated = false;
+		cy.window().then(() => {
+			this.swup.options.linkToSelf = 'navigate';
+			this.swup.hooks.once('visit:start', () => (navigated = true));
+		});
+		cy.scrollTo(0, 200);
+		cy.window().its('scrollY').should('equal', 200);
+		cy.get('[data-cy=nav-link-self]').click();
+		cy.window().its('scrollY').should('equal', 0);
+		cy.window().should(() => {
+			expect(navigated).to.be.true;
+		});
+	});
 });
 
 describe('Redirects', function () {
--- a/cypress/e2e/main.spec.js
+++ b/cypress/e2e/main.spec.js
@@ -324,6 +324,7 @@ describe('Navigation', function () {
 describe('Link resolution', function () {
 	beforeEach(() => {
 		cy.visit('/link-resolution.html');
+		cy.wrapSwupInstance();
 	});
 
 	it('should skip links to different origins', function () {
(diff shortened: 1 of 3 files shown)


daun:
@hirasso I don't think this is ready to merge. I mainly wanted to start a discussion here about how these changes would impact `resolveUrl` and how we can support both going forward.

hirasso:
Could we even have a callback `({ url, trigger }) => 'scroll' | 'navigate' | 'reload'`, returning one of the three options based on the current URL? I'm thinking of the issue another user raised with submitting a form to the same URL. They might want another behavior for other links.

Regarding `resolveUrl`: I think as long as all this happens before `isEqualResolvedUrl` is being called, everything should be fine. The rule should be that equal resolved URLs should never fire a `visit:self` event. They should just do nothing.

daun:
Callback sounds interesting! In that case, `ignoreVisit` and `linkToSelf` seem to have quite the overlap, no? They both manage what happens for a given URL — one does it for all URLs, the other does it for identical URLs. There might be a good way of unifying this, but I'm failing to see it at the moment.

> I'm thinking of the issue another user raised with submitting a form to the same URL. They might want another behavior for other links.

If we're thinking of the same [issue](https://github.com/swup/swup/issues/735), it was about allowing link elements with a URL to the form page itself as a 'hacky' form reset. That case would be covered by a simple string option as well.

How about doing this on-demand? Leave it as a string option for now, and add the callback if we see a concrete need or it gets requested.

>  I think as long as all this happens before `isEqualResolvedUrl` is being called, everything should be fine.

Ah, I see. So it's okay if `link:self` fires for actually identical URLs, and short-circuits even before there's a chance of resolving URLs and comparing those? Would setting the new option to `navigate` break sites that rely on `resolveUrl` for client-side-only filter-type situations? Correct me if I'm wrong here.

hirasso:
Thanks for the issue link!

> When users are handling forms, it would be elegant to allow **both POSTing the same URL** and clearing of form data by just visiting a link to the form.

That reads to me as if they also want to POST a form to the same URL. They might still want to keep swup's default behavior of scrolling to the top for other links. 

Your thoughts about the overlap of `linkToSelf` and `ignoreVisit` sound very reasonable. To be honest, I had to re-visit the [`ignoreVisit` docs](https://swup.js.org/options/#ignore-visit) to figure out what exactly would happen if returning `true`. The browser would do a hard reload! So the `'reload'` part should already be solvable by giving links to the same page a `[data-no-swup]` attribute or using the callback:

```js
{
  ignoreVisit: (url, { el, event } = {}) => {
    if (url === swup.getCurrentUrl()) return true; // reload for links to the same page
    return el?.closest('[data-no-swup]');
}
```

Based on that we could maybe simplify the option `linkToSelf` to a callback as well:

```
{
  linkToSelf: (url, { el, event } = {}) => {
    if (url === myFormsUrl) return 'navigate';
    return 'scroll'
}
```

I'm also a bit stumped by the ideal API solution here. We should maybe have a call about it.

daun:
@hirasso I've added support for passing a callback function to `linkToSelf`. Also, added a conditional to not update or create a history visit if the current and next URLs are identical.

hirasso (approved):
Looks good to me!

PR #757, opened 2023-08-08
Feature/overwrite scroll target

hirasso:
**Description**

Right now, if modifying `visit.scroll.target` on the fly like this:

```js
swup.hooks.once('visit:start', (visit) => (visit.scroll.target = '#anchor'));
```

`#anchor` will be appended to the URL of the target page. It's debatable if we want that or not. 

This PR consists only of a failing test for the expectation that the hash _shouldn't_ be appended. Opening a draft PR to start the discussion.

**Checks**

- [x] The PR is submitted to the `master` branch
- [x] The code was linted before pushing (`npm run lint`)
- [ ] All tests are passing (`npm run test`)
- [x] New or updated tests are included

PR #758, opened 2023-08-08
Don't append the hash if setting `visit.scroll.target` on the fly

hirasso:
Replaces #757 

**Description**

- Uses `visit.to.url + visit.to.hash` for creating history entries
- Ignores `visit.scroll.target` for the history URL
- Scrolls to either `visit.scroll.target` or `visit.to.hash`

**Checks**


hirasso:
Updated the behavior so that only the hash of the trigger makes it into the history, while a modified `visit.scroll.target` will be ignored when creating/updating history records.

daun (approved):

PR #762, opened 2023-08-10
Set up ESLint

daun:
**Description**

- Set up eslint + typescript-eslint
- Apply fixes for most rules
- Add eslint check to `lint` npm script to check in PRs

**Next steps**

- Enable stricter `@typescript-eslint/recommended-type-checked` preset
- Add better types to Hook handlers (currently `any`)
- Extract eslint config into shared repo?
- Apply eslint to plugins as well?

**Checks**


.eslintrc.cjs
```diff
@@ -0,0 +1,16 @@
...
+		'no-unused-vars': ['error', { vars: 'all', args: 'none' }],
+		'@typescript-eslint/ban-ts-comment': ['error', { 'ts-ignore': 'allow-with-description' }],
+	},
+	parser: '@typescript-eslint/parser',
+	parserOptions: {
+		project: true,
+		tsconfigRootDir: __dirname
+	},
+	root: true
+};
```
hirasso, line 16: Could we make this re-usable for plugins somehow?

daun: We should! I think eslint configs can live in repos as well.

src/modules/Cache.ts
```diff
@@ -53,7 +53,7 @@ export class Cache {
 	}
 
 	/** Update a cache record, overwriting or adding custom data. */
-	update(url: string, payload: Record<string, any>) {
+	update(url: string, payload: object) {
```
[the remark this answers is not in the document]
daun: It didn't throw, but it wouldn't allow passing in data without a string index signature, which is super annoying. Object seems to work for arbitrary data (see the tests where we could remove the type casting completely).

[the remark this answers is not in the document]
hirasso: Nice!

.eslintrc.cjs
daun: I'll take care of it.

daun:
@hirasso Extracting the ESLint config into its own repo is messier than expected; I'll keep that as a to-do for a rainy day.

PR #764, opened 2023-08-11
Cache control per visit

### LINKED ISSUE #763: Allow bypassing cache for load and save
**Describe the problem**

There should be a way of disabling the cache to help with forms, especially on POST requests.

1. Bypass cache on load: ignore any previously cached pages
2. Skip cache on save: do not store the response in the cache

1. Extend `FetchOptions` with two parameters, e.g.
  a. `bypassCache` and `skipCacheSave` with default `false`
  b. `getFromCache` and `saveToCache` with default `true`
3. Extend `FetchOptions` with a `cacheControl` parameter that mimics the [Cache-Control header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control?retiredLocale=de#directives)
4. Directly read from the request headers and interpret the `Cache-Control` header ourselves

**Alternatives considered**

- Manually clearing the cache before a request
- Manually clearing the cache after a request

- [ ] Nice to have
- [x] Would make my life a lot easier

daun:
**Description**

- Store cache settings in visit object: `{ read: true, write: true }`
- Add cache option to `swup.navigate`:  `{ read?: true, write?: true }`
- Only cache responses of `GET` requests
- Increases bundle size by 100 bytes

**Drive-by changes**

- Remove `this.cache.clear()` from `renderPage` as it was meant for Preload Plugin which now requires the cache

**Checks**

**Additional information**

Closes #763

### DIFF
--- a/cypress/e2e/main.spec.js
+++ b/cypress/e2e/main.spec.js
@@ -103,19 +103,19 @@ describe('Cache', function () {
 
 	it('should cache pages', function () {
 		this.swup.navigate('/page-2.html');
-		cy.shouldBeAtPage('/page-2.html');
+		cy.shouldBeAtPage('/page-2.html', 'Page 2');
 		cy.shouldHaveCacheEntry('/page-2.html');
 	});
 
 	it('should cache pages from absolute URLs', function () {
 		this.swup.navigate(`${baseUrl}/page-2.html`);
-		cy.shouldBeAtPage('/page-2.html');
+		cy.shouldBeAtPage('/page-2.html', 'Page 2');
 		cy.shouldHaveCacheEntry('/page-2.html');
 	});
 
 	it('should not cache pages for POST requests', function () {
 		this.swup.navigate(`${baseUrl}/page-2.html`, { method: 'POST' });
-		cy.shouldBeAtPage('/page-2.html');
+		cy.shouldBeAtPage('/page-2.html', 'Page 2');
 		cy.shouldNotHaveCacheEntry('/page-2.html');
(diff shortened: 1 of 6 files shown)


hirasso:
Amazing!

I stumbled a bit upon the boolean for both behaviours. Without knowing the internals, I would think that this would only disable loading from the cache. I wouldn't think about saving to the cache.

daun:
@hirasso Do you think a boolean false should disable reads but keep writes enabled?

hirasso:
I think I would force users to think about what they want and remove the wildcard boolean altogether

daun:
I've removed the boolean option. Since we need to sanitize the value to an object during runtime anyway, technically booleans still work the same, however they'll now throw an error in TS projects.

daun:
I've pushed a change to only cache `GET` requests in general to align swup with browser behavior.

daun:
The tests are really flaky here; they work locally every time but fail intermittently in CI. Will need debug them.

daun:
Pushed a fix for the flaky tests — seems to be working now.

hirasso:
Huh? How did you come up with that fix?! Can you explain why this makes the tests less flaky?😵‍💫😄

daun:
We were only checking if the new URL was matching the expected page, but that could match even before the hooks have run since the URL gets updated before any requests are sent. Now we're making sure the content was updated as well before continuing.

hirasso (approved):

hirasso:
Interesting!

PR #768, opened 2023-08-16
Fix scroll restoration

daun:
Fix two issues with scrolling.

**1: Scroll restoration**

Make sure the browser doesn't try to restore scroll positions when animating history visits by conditionally setting `history.scrollRestoration = 'manual'`. This was already in the Scroll Plugin, but is required in swup as well to avoid a flash of scrolled-down content before swup resets it to the top.

Before:

https://github.com/swup/swup/assets/22225348/1ff34a7a-9d07-4cd1-9ea9-aefb3ef66dfb

After:

https://github.com/swup/swup/assets/22225348/6ff8b0bd-65e6-43e4-b007-2d9b30a9cf15

**2: Scroll targets on history visits**

The recent addition of `visit.to.hash` had a flaw on history visits: when going back to a page with a `#hash` in the URL, the browser would first restore the previous scroll position, then scroll to the anchor element. Solved by allowing boolean `false` in `visit.scroll.target` and disabling it in the popstate handler. The scroll target will now be ignored on history visits.

**Other changes**

Simplified the surface of the object passed into `createVisit` — most keys are only needed once, so it makes more sense to keep the most common keys (from, to, hash, el, event) and just set any others directly after creating the visit by doing `this.visit.x = y`. Not breaking since only used internally and in no plugins.

**Checks**

### DIFF
--- a/src/Swup.ts
+++ b/src/Swup.ts
@@ -331,6 +331,7 @@ export default class Swup {
 		// Disable animation & scrolling for history visits
 		this.visit.animation.animate = false;
 		this.visit.scroll.reset = false;
+		this.visit.scroll.target = false;
 
 		// Animated history visit: re-enable animation & scroll reset
 		if (this.options.animateHistoryBrowsing) {
--- a/src/Swup.ts
+++ b/src/Swup.ts
@@ -315,18 +315,10 @@ export default class Swup {
 		}
 
 		const { url, hash } = Location.fromUrl(href);
-		const animate = this.options.animateHistoryBrowsing;
-		const resetScroll = this.options.animateHistoryBrowsing;
 
-		this.visit = this.createVisit({
-			to: url,
-			hash,
-			event,
-			animate,
-			resetScroll
-		});
+		this.visit = this.createVisit({ to: url, hash, event });
 
-		// Mark as popstate visit
+		// Mark as history visit
 		this.visit.history.popstate = true;
 
 		// Determine direction of history visit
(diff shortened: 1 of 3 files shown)


src/modules/Visit.ts
```diff
@@ -52,7 +52,7 @@ export interface VisitScroll {
 	/** Whether to reset the scroll position after the visit. Default: `true` */
 	reset: boolean;
 	/** Anchor element to scroll to on the next page. */
-	target?: string;
+	target?: string | false;
```
daun, line 55: Allow boolean false to disable scroll-to-target.

src/modules/scrollToContent.ts
```diff
@@ -7,7 +7,7 @@ import Swup from '../Swup.js';
 export const scrollToContent = function (this: Swup): boolean {
 	const options: ScrollIntoViewOptions = { behavior: 'auto' };
 	const { target, reset } = this.visit.scroll;
-	const scrollTarget = target || this.visit.to.hash;
+	const scrollTarget = target ?? this.visit.to.hash;
```
daun, line 10: Allow distinguishing between `undefined` scroll target (= check for hash) and `false` scroll target (= disable scroll-to-target feature, ignore hash).

src/Swup.ts
```diff
@@ -331,6 +328,17 @@ export default class Swup {
 			this.visit.history.direction = direction;
 		}
 
+		// Disable animation & scrolling for history visits
+		this.visit.animation.animate = false;
```
daun, line 332: Preferred the verbose, repetitive setting of visit keys here for clarity.

src/modules/Visit.ts
```diff
@@ -82,38 +82,23 @@ export interface VisitInitOptions {
...
-		from = this.currentPageUrl,
-		hash,
-		animate = true,
-		animation: name,
-		el,
-		event,
-		action = 'push',
-		resetScroll: reset = true
-	}: VisitInitOptions
+	{ to, from = this.currentPageUrl, hash, el, event }: VisitInitOptions
```
daun, line 92: Simplify visit init object to only the most common keys.

hirasso (approved):
Great catches!!

PR #772, opened 2023-08-20
Update history on links to self

daun:
**Description**

- Clicking a link to anchor on the current page will update the visible URL
- Clicking a link to the current page without anchor will not update the visible URL
- This PR makes both `link:self` and `link:anchor` update the URL

**Checks**

### DIFF
--- a/src/Swup.ts
+++ b/src/Swup.ts
@@ -284,6 +284,7 @@ export default class Swup {
 								return this.performNavigation();
 							case 'scroll':
 							default:
+								updateHistoryRecord(url);
 								return this.scrollToContent();
 						}
 					});


hirasso:
I think I'm not in favor of updating the URL if `link:self` doesn't perform a navigation but only scrolls to the top.

Nothing would happen if people would click the back button after such a pushState, right?

daun:
@hirasso It only updates the current history entry, it doesn't push a new one. Same as clicking an anchor, which only updates the history entry and doesn't push a new one. I'd say these should act the same way.

daun:
@hirasso The scenario this is needed is emulating the `:target` pseudo-class: if we don't update the URL when clicking a link without anchor, we can't remove the emulate `[data-swup-scroll-target]` attribute when clicking on a link that would change the target element to nothing, i.e. it has no hash.

hirasso:
Ah I see! Nice 😻

hirasso (approved):

PR #774, opened 2023-08-21
Fix history index

daun:
**Description**

- While experimenting with parallel layouts and history stacks, I noticed the history index isn't set correctly
- The value was never permanently incremented, leading to index `2` for all history states 🙃
- This PR adds a few things around that topic:
  - Read index from existing history state on load
  - Only create initial history index if none exists
  - Actually increment the index
  - Only set direction if an index was found

**Checks**

### DIFF
--- a/src/Swup.ts
+++ b/src/Swup.ts
@@ -328,7 +328,7 @@ export default class Swup {
 
 		// Determine direction of history visit
 		const index = (event.state as HistoryState)?.index ?? 0;
-		if (index) {
+		if (index && index !== this.currentHistoryIndex) {
 			const direction = index - this.currentHistoryIndex > 0 ? 'forwards' : 'backwards';
 			this.visit.history.direction = direction;
 			this.currentHistoryIndex = index;
--- a/src/Swup.ts
+++ b/src/Swup.ts
@@ -183,8 +183,10 @@ export default class Swup {
 		// Mount plugins
 		this.options.plugins.forEach((plugin) => this.use(plugin));
 
-		// Modify initial history record
-		updateHistoryRecord(null, { index: 1 });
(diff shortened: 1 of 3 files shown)


src/Swup.ts
```diff
@@ -181,8 +183,10 @@ export default class Swup {
 		// Mount plugins
 		this.options.plugins.forEach((plugin) => this.use(plugin));
 
-		// Modify initial history record
-		updateHistoryRecord(null, { index: 1 });
+		// Create initial history record
+		if ((history.state as HistoryState)?.source !== 'swup') {
```
daun, line 187: Only create initial history index if none exists

src/Swup.ts
```diff
@@ -323,10 +327,11 @@ export default class Swup {
 		this.visit.history.popstate = true;
 
 		// Determine direction of history visit
-		const index = Number((event.state as HistoryState)?.index);
-		if (index) {
+		const index = (event.state as HistoryState)?.index ?? 0;
```
daun, line 330: Only set direction if an index was found

src/modules/navigate.ts
```diff
@@ -123,8 +123,8 @@ export async function performNavigation(
 			) {
 				updateHistoryRecord(newUrl);
 			} else {
-				const index = this.currentHistoryIndex + 1;
-				createHistoryRecord(newUrl, { index });
+				this.currentHistoryIndex++;
```
daun, line 126: Actually increment the index

hirasso:
Are there tests now for the index?

daun:
@hirasso Yes, there are!

hirasso (approved):

PR #776, opened 2023-08-24
Add unique id to visit object

daun:
**Description**

- Add a unique random `visit.id` when creating new visits
- Ignore outdated visits if IDs mismatch
- As much as possible, refer to local `visit` var instead of `this.visit` to avoid modifying newer visit
- Add tests for the shape of the Visit object
- Add tests for outdated visits being ignored

**Goals**

- Identify and ignore outdated visits with more certainty
- Allow checking if we're still in the same visit in `hooks.once()` calls

**Allow identifying outdated visits**

Currently, we're checking if a newer visit was started in the meantime by comparing the two requested URLs. That doesn't work for the now-possible same-link visits, and it's also safer and clearer to just compare IDs:

```diff
- if (!this.isSameResolvedUrl(getCurrentUrl(), requestedUrl)) return;
+ if (visit.id !== this.visit.id) return;
```

**Allow checking for same visit in `once` hooks**

Currently, when registering a hook `once`, there is no guarantee that the run hook belongs to the same visit, e.g.:

- When clicking links twice in quick succession, it's possible for the second link to fetch earlier
- When hooking into animation functions that might be skipped, we'd get the animation of the **next** visit

By having a unique id on the visit object we can check in `once` hooks if we're indeed dealing with the same visit:

```js
swup.hooks.on('visit:start', (visit) => {
  if (someCondition) {
    swup.hooks.once('animation:in:await', ({ id }) => {
      if (visit.id !== id) return;
      await someCustomFunction();
    }
  }
});
```

**Checks**

### DIFF
--- a/cypress/e2e/main.spec.js
+++ b/cypress/e2e/main.spec.js
@@ -389,6 +389,31 @@ describe('Navigation', function () {
 		cy.shouldHaveH1('Page 2');
 	});
 
+	it('should ignore visit if a new visit has started', function() {
+		cy.delayRequest('/page-2.html', 1000);
+		cy.triggerClickOnLink('/page-2.html');
+		cy.wait(50);
+		cy.triggerClickOnLink('/page-3.html');
+		cy.shouldBeAtPage('/page-3.html', 'Page 3');
+		cy.wait(1000);
+		cy.shouldBeAtPage('/page-3.html', 'Page 3');
+	});
+
+	it('should ignore visit if a new visit to same URL has started', function() {
+		const titles = [];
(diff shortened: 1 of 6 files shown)


hirasso:
Keeping this id independent of the history.state random key is exactly what we want here, right?

Every visit will have it's unique id, no matter if navigating through links or the history stack, right?

daun:
@hirasso Exactly, the idea is to identify visits as isolated actions, regardless of source. For the history, we can differentiate those with the (now working 😅) index property of each history state object.

PR #779, opened 2023-08-30
Prepare ViewTransition support

daun:
**Description**

- Trying to get swup to play nice with the new [ViewTransitions API](https://developer.mozilla.org/en-US/docs/Web/API/ViewTransition)
- Turns out the only (!) thing required to get them working is wrapping all DOM changes in a new hook
- We can now replace that hook and wrap the DOM changes with a `document.startViewTransition()` call
- That's all — ViewTransition are now supported 🤠

**Example**

The animations are all done in CSS using `::view-transition-*` pseudo-elements. Parallel animations without plugins!

https://github.com/swup/swup/assets/22225348/2f716aca-1d6f-4606-9ab9-4a828ec97b5f

**Checks**

### DIFF
--- a/src/modules/Hooks.ts
+++ b/src/modules/Hooks.ts
@@ -31,6 +31,7 @@ export interface HookDefinitions {
 	'scroll:top': { options: ScrollIntoViewOptions };
 	'scroll:anchor': { hash: string; options: ScrollIntoViewOptions };
 	'visit:start': undefined;
+	'visit:transition': undefined;
 	'visit:end': undefined;
 }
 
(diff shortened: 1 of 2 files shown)


hirasso:
😻 Beautiful!!! Should we add tests for this? Anything that makes sense? I don't think so :)

hirasso (approved):

hirasso:
Yes let's do this! It also feels to me like view transitions support should be a core feature. Call sounds good! 📞

hirasso:
Sounds like a reasonable plan! 👌

daun:
🚀

PR #792, opened 2023-09-22
fix: ignore double-clicks on links

### LINKED ISSUE #791: Triggering link:click twice seems to interfere with content:replace and animation:in
### Discussed in https://github.com/orgs/swup/discussions/790

<div type='discussions-op-text'>

<sup>Originally posted by **VisiaMedia** September 21, 2023</sup>
Hi there,

First off, thank you for the amazing library and extensive documentation. Implementing Swup into my project has been quite a ride, but I'm loving it so far.

After upgrading V3 to V4, I've been running into some unexpected behavior. Which I suspect is caused by the JS Plugin:

Navigating between pages works fine in most cases, but when (accidentally) double-clicking, the second link:click event seems to stop the expected sequence of events, which causes animation:in to not fire at all. content:replace then won't fire and the transition stays stuck between the out and in stages.

I installed the Debug Plugin to find out what exactly happens. Here's what I ran in to:

**Regular link:click event:**
Screenshot 2023-09-21 at 16 45 44

**Double-clicking:**
Screenshot 2023-09-21 at 16 45 59

I'm not 100% sure if this behavior is caused by the JS Plugin, and at this point my project relies too much on it to remove the plugin. As my project is quite large, I can't extract a minimal demo. But I do have a build which you can check out [here](https://visiamedia.netlify.app/).

I hope someone can point me into the right direction. Thanks in advance!</div>

hirasso:
Fixes #791

**Description**

Ignores clicks on links if swup is currently already navigating towards the link's url. For that to work, we'll need some way to detect if a swup visit has settled. I've added a top level `visit.settled` for that.

**Questions**

- Where exactly would we want to set `settled` to true, before, after or inside `visit:end`?
- Naming: should it actually be called `visit.ended`, to comply with `visit:end`?

**Drive-by**

- Declare globals on `Window` in tests
- Fix test for disabling cache completely using navigation options

**Checks**

- [ ] ~The documentation was updated as required~

**Additional information**

### DIFF
--- a/src/Swup.ts
+++ b/src/Swup.ts
@@ -96,6 +96,8 @@ export default class Swup {
 	protected currentHistoryIndex: number;
 	/** Delegated event subscription handle */
 	protected clickDelegate?: DelegateEventUnsubscribe;
+	/** Navigation status */
+	navigating: boolean = false;
 
 	/** Install a plugin */
 	use = use;
@@ -145,7 +147,7 @@ export default class Swup {
 		this.cache = new Cache(this);
 		this.classes = new Classes(this);
 		this.hooks = new Hooks(this);
-		this.visit = this.createVisit({ to: '', settled: true });
+		this.visit = this.createVisit({ to: '' });
 
 		this.currentHistoryIndex = (history.state as HistoryState)?.index ?? 1;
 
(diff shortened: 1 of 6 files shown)


hirasso:
Actually, while implementing this I'm getting a feeling that this shouldn't be `visit.settled`, but rather a top level property in swup itself `navigating` or something. Would make the code a lot simpler I imagine.

daun:
Let's get on a call to figure this out 👽 Easier to go through the scenarios.

hirasso:
...tests are passing now! phew, not so easy to figure out playwright's `evaluate` function :)

daun (approved):
🦑

PR #793, opened 2023-09-22
fix: types in tests

hirasso:
**Description**

My IDE didn't pick up the globals declared in `main.spec.ts`. Instead, it started working when I added the globals declaration to [`./tests/support/swup.ts`](https://github.com/swup/swup/blob/1d2df420c6608464fae507a58af71ab3207f39cd/tests/support/swup.ts#L6-L13), I suspect because that file is being imported by the test files.

**Drive-By**

- Fix typings in test for navigation without cache
- Add new global type `navigated: () => boolean` for link resolution tests
- Format test files

**Checks**

### DIFF
--- a/tests/functional/cache.spec.ts
+++ b/tests/functional/cache.spec.ts
@@ -1,7 +1,11 @@
 import { expect, test } from '@playwright/test';
 
 import { expectToBeAt } from '../support/commands.js';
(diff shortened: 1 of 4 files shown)


daun (approved):
Not sure if this needs more work, approving just in case :)

PR #797, opened 2023-09-26
Mangle private properties

daun:
**Description**

- Decrease bundle size by mangling private properties: `swup._performNavigation` → `swup.d`
- Requires prepending underscores for now
- Would prefer if this could be automatic just by the `protected` status of a property/method
- Tests currently breaking as we're relying on access to some private props of internal classes

**Checks**

### DIFF
--- a/mangle.json
+++ b/mangle.json
@@ -1,0 +1,9 @@
+{
+  "minify": {
+    "mangle": {
+      "properties": {
+        "regex": "^_"
+      }
+    }
+  }
+}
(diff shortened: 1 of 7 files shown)


hirasso:
I wouldn't do it if it only works with underscores. I never liked these, since I first saw them used in Flash files 🥸

daun:
Yeah, same. If I find a way to do this automatically for all `protected` members, I'll revive this.

PR #798, opened 2023-09-26
Ensure typescript 5 is available

hirasso:
**Description**

Recent research shows that we only had typescript 5 available by accident. Let's change that and make it explicit.

**Checks**


daun (approved):
👨‍🔧

PR #802, opened 2023-10-02
Native mode for built-in View Transition support

daun:
**Description**

- Builds on #779 
- New `native` option to enable [View Transition](https://developer.mozilla.org/en-US/docs/Web/API/View_Transitions_API) support
- Native mode skips all `animation:*` hooks
- Works great in local playground for basic scenarios

**To-do**

- Fallback strategy for unsupported browsers
- Strategy for dynamic containers
- Functional tests

**Questions**

- How do we deal with dynamic containers? How can we mark them as "changing"?
  - We can use a new class `is-transitioning` to mark containers
  - We can replace the whole body as users now have precise control over what is transitioned
  - We can make users rely on custom routes and fragment visit names to specify view transitions
- Should the `animation:*` hooks fire on native transitions as well? I'd say no, as they're tied to swup's CSS logic

**Follow-ups**

- We should think about the concept of pluggable/modular animation handlers (Native, CSS, JS, None)
- Something like that would reduce bundle size and improve control flow
- Sounds like a project for v5 🤪
- Or we integrate this into v4 and load a CSS handler by default to avoid breaking changes

**Checks**

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,6 +19,7 @@
         "@playwright/test": "^1.37.1",
         "@swup/browserslist-config": "^1.0.0",
         "@swup/prettier-config": "^1.0.0",
+        "@types/dom-view-transitions": "^1.0.2",
         "@types/jsdom": "^21.1.1",
         "@typescript-eslint/eslint-plugin": "^6.3.0",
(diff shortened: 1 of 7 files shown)


daun:
Closing this in favor of #837 for now.

PR #819, opened 2023-11-14
Use official bundle size badge

daun:
**Description**

- Use official [shields.io badge](https://shields.io/badges/npm-package-minimized-gzipped-size-select-exports) for getting package size via bundlejs
- Remove previous dodgy json access version


hirasso (approved):
👾

Issue #811, opened 2023-10-29
Think about chances for optimizations when clicking a link during a visit


hirasso:
**Describe the problem**

[This discussion](https://github.com/orgs/swup/discussions/809) brought it up one more time: Clicking another link during a running page visit [now works](https://github.com/swup/swup/pull/464), but breaks currently running animations. I'd like to get together in a call sometime to think about possible better solutions than what we have now.

**Describe the proposed solution**

<del>

Very rough idea, just as a starting point:

1. Create some kind of "parent" for each page visit, for example a class `VisitController`
2. That controller could have a method `abort` or `disconnect` that would be called each time a new visit was triggered
3. When disconnected, the controller:
    - wouldn't trigger hooks
    - wouldn't alter the dom
    - would completely destroy itself as soon as it can

</del>

The above doesn't make sense at all. I don't know what I was thinking 🤡 

**Notes**

- would this even work with the css-based animation logic of swup?
- if we find out that it could actually work better, could we utilize the existing `Visit` object for it?

**How important is this feature to you?**

- [x] Nice to have. I'd just like to rethink it and document our thoughts about it.

daun:
Sounds like a great topic to explore on a call sometime 🔮

daun:
I guess we could always expand [the current checks](https://github.com/swup/swup/blob/master/src/modules/navigate.ts#L149C1-L152C5) we're doing to identify and ignore visits that have been superseded by a new one in the meantime.

```ts
// Abort if another visit was started in the meantime
if (visit.id !== this.visit.id) {
  return false;
}
```

hirasso:
That will be one part of the solution for sure – I'm unsure about these two:

1. how do we cleanly cancel/restart CSS animations? Something like 
    1. remove animation classes
    2. wait one tick
    3. initialize the next visit
2. how would we go about it if the user clicks another link during the **out**-animation?

Does animating still make sense in this scenario at all, if we think about special animations for routes (Route Name Plugin, JS Plugin)? My initial thought was that the cleanest approach would be to cancel eventually running animations alltogether.

daun:
Regarding 1, that sounds perfectly reasonable. 

On point 2, that's already solved, at least when using CSS class-based transitions. The above check `visit.id !== this.visit.id` will effectively just drop any visits if during the out-animation another visit was requested. Adding the css class for the out-animation again will not really do anything. For clicks during the in-animation, this will not yet work, as well as for JS animations, as these do get repeated, I think.

[...]

PR #825, opened 2023-11-16
feat: handle rapid navigation

hirasso:
Related: #811 

**Description**

Optimizes behavior on rapid navigation.

**Checks**

**Additional information**

### DIFF
--- a/src/modules/Classes.ts
+++ b/src/modules/Classes.ts
@@ -3,7 +3,14 @@ import { queryAll } from '../utils.js';
 
 export class Classes {
 	protected swup: Swup;
-	protected swupClasses = ['to-', 'is-changing', 'is-rendering', 'is-popstate', 'is-animating'];
+	protected swupClasses = [
+		'to-',
+		'is-changing',
+		'is-rendering',
+		'is-popstate',
+		'is-animating',
+		'is-leaving'
+	];
 
 	constructor(swup: Swup) {
 		this.swup = swup;
@@ -37,6 +44,7 @@ export class Classes {
 	clear(): void {
 		this.targets.forEach((target) => {
 			const remove = target.className.split(' ').filter((c) => this.isSwupClass(c));
+			console.log(...remove);
 			target.classList.remove(...remove);
 		});
 	}
--- a/src/modules/Classes.ts
+++ b/src/modules/Classes.ts
@@ -44,7 +44,6 @@ export class Classes {
 	clear(): void {
 		this.targets.forEach((target) => {
 			const remove = target.className.split(' ').filter((c) => this.isSwupClass(c));
-			console.log(...remove);
 			target.classList.remove(...remove);
 		});
 	}
--- a/src/modules/Hooks.ts
+++ b/src/modules/Hooks.ts
@@ -350,7 +350,7 @@ export class Hooks {
 			await this.run(before, args);
 		}
 
-		const [result] = await this.run(visitId === this.swup.visit.id ? handler : [], args);
+		const [result] = await this.run(handler, args);
 
 		if (visitId === this.swup.visit.id) {
 			await this.run(after, args);
--- a/src/modules/Hooks.ts
+++ b/src/modules/Hooks.ts
@@ -342,11 +342,18 @@ export class Hooks {
 		args: HookArguments<T>,
 		defaultHandler?: HookDefaultHandler<T>
 	): Promise<Awaited<ReturnType<HookDefaultHandler<T>>>> {
+		const visitId = this.swup.visit.id;
+		// @TODO: how can we return early here?
+
 		const { before, handler, after } = this.getHandlers(hook, defaultHandler);
 		await this.run(before, args);
 		const [result] = await this.run(handler, args);
 		await this.run(after, args);
-		this.dispatchDomEvent(hook, args);
+
+		if (visitId === this.swup.visit.id) {
+			this.dispatchDomEvent(hook, args);
+		}
+
 		return result;
 	}
 
--- a/src/modules/Hooks.ts
+++ b/src/modules/Hooks.ts
@@ -343,12 +343,19 @@ export class Hooks {
 		defaultHandler?: HookDefaultHandler<T>
 	): Promise<Awaited<ReturnType<HookDefaultHandler<T>>>> {
 		const visitId = this.swup.visit.id;
-		// @TODO: how can we return early here?
 
 		const { before, handler, after } = this.getHandlers(hook, defaultHandler);
-		await this.run(before, args);
+
+		if (visitId === this.swup.visit.id) {
+			await this.run(before, args);
+		}
+
+		// @TODO: how can we return early here without executing the hook if the visit has expired?
 		const [result] = await this.run(handler, args);
-		await this.run(after, args);
+
+		if (visitId === this.swup.visit.id) {
+			await this.run(after, args);
+		}
 
 		if (visitId === this.swup.visit.id) {
 			this.dispatchDomEvent(hook, args);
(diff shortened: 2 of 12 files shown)


hirasso:
## Before (fickers when navigating during a transition)


https://github.com/swup/swup/assets/869813/a24f90c6-9a28-4b28-9047-f12b7b74955f

## After


https://github.com/swup/swup/assets/869813/144f0fb7-9f65-4231-84ea-1f79347e2068

hirasso:
The hooks logic is beyond me... It's quite possible that I'm doing checks for `visitId === this.visit.id` a few times too often 😅 @daun feel free to remove them as you see fit (or tell me where the checks don't make sense).

It could also be a better idea to do something like this in `createVisit()`: 

```js
this.visit.expired = true;
```

...then we could check for that instead. In any of the functions: 

```js
const visit = this.visit;
//...later, after async code:
if (visit.expired) return;
```

What do you think?

daun:
Great idea. I like having a separate flag for this, much cleaner. Other possible names could be `stale` or `canceled`.

Thinking slightly bigger, we could call it `ignore` and re-use that to allow ignoring visits from hooks in general. Maybe with a shortcut function `ignore()` that would set `visit.ignore` to `true`. Nice API inside hooks: `(visit) => visit.ignore()`.

There's an argument to be made for keeping the two concepts (expired and ignored) separate, though.

hirasso:
I'd like to keep it as simple as possible for now – it's overwhelming enough already 😄 I'll try my luck with the `expired` flag, let's see where it takes us.

daun:
Makes sense to stick with one task here 👽

daun:
One thing I don't feel very good about is that the `Hooks` class now needs to know about the concept of a visit and checks for `this.swup.visit.expired`. Aren't we checking inside the hook handlers anyway and returning early? In that case, those checks shouldn't really be doing anything.

hirasso:
OK, I'd say `visit.expired` is working great! Regarding your idea with `visit.ignore` and some related API: I always find `ignore` to be much too vague of a description. *What* should be ignored? Also, don't we already have it in swup, leading to a hard page load? I thought so.

Expiring a visit is already possible right now, since I'm feeding the correct visit to `hooks.call` now. So something like this already works. I'm using to test the functionality:

https://github.com/swup/swup/blob/fd11f443b1e9f941bb85104ab8a913cc412298ad/tests/functional/rapid-navigation.spec.ts#L43-L46

We shouldn't advertise this, though, since I don't see how it could be helping consumers instead of just confusing them. Swup will **stop proceeding with the visit** as soon as it detects that a visit has `expired`. Blank page ;)

hirasso:
Ready from my end :)

daun:
As it's part of the `Visit` interface, we're kind of advertising it. I think it's a useful feature we should expose to users, it's just that `expired` doesn't work too well then as a name then. Calling it `canceled` could cover both cases, I guess? Swup can `cancel` the current visit when a new one starts, and users can `cancel` a visit inside a hook if they feel the need to. WDYT?

hirasso:
I don't care too much. Will call it canceled. Or `stopped`? 😄

daun:
For me, `canceled` works for both use cases and seems clearest for users. Slightly better than `stopped`.

hirasso:
ok. Last comment before weekend: In this special case, maybe prefixing the property with an underscore would be fitting. It really is an internal API. So `visit._cancelled`?

daun:
I really dislike underscores aesthetically and we don't use them anywhere, so I'd be in favor of avoiding them. I don't think it's necessary here — making this public would allow canceling visits from the consumer side, which is a nice feature I think.

hirasso:
Ok. I never liked them either, but it's a widely used convention. But again, no strong opinion. Will rename the darn thing to `cancelled` 😄

hirasso:
Just as a proof of concept I marked `visit.cancelled` as `@private` in the last commit. It's working! The `Visit` interface when I build swup:

```typescript
export interface Visit {
    /** A unique ID to identify this visit */
    id: number;
    /** The previous page, about to leave */
    from: VisitFrom;
    /** The next page, about to enter */
    to: VisitTo;
    /** The content containers, about to be replaced */
    containers: Options['containers'];
    /** Information about animated page transitions */
    animation: VisitAnimation;
    /** What triggered this visit */
    trigger: VisitTrigger;
    /** Cache behavior for this visit */
    cache: VisitCache;
    /** Browser history behavior on this visit */
    history: VisitHistory;
    /** Scroll behavior on this visit */
    scroll: VisitScroll;
    // <-- No cancelled here :)
}
```
Will revert it if we decide that we have a good use case for `visit.cancelled`.

daun:
😍

daun:
Leaving a suggestion here for discussion on monday: we could/should introduce a `visit:cancel` hook that is run whenever a visit is "cut short" in favor of a newer one started in the meantime. My projects (and I assume a few plugins as well) rely on `visit:start` to set temporary state and on `visit:end` to undo it, e.g. Parallel Plugin to reset containers when a visit finishes:

```ts
this.on('visit:end', this.cleanupContainers);
```

Having `visit:cancel` as well would allow responding differently to these two very different scenarios.

daun:
A closely related point: with this change, we probably need to do an extra round of testing with all plugins to make sure they keep working as intended. A few possible cases that make use of `visit:end` are A11y Plugin, Fragment Plugin, Parallel Plugin, and (maybe?) Progress Plugin.

daun:
Admittedly, parallel animations will never look amazing even with this change, as they're not cleared by the code in question.

daun:
Which brings up another point for discussion: having an option `cancelRunningVisits` would allow us to get rid of a lot of this logic by always waiting for `visit:end` before triggering a new visit. The first advantage is reduced complexity, the second advantage is consistency/future-proofing: this would resemble how the native ViewTransition API handles things by making the page `inert` and not allowing interactions during animations. Not quite the same technically, but in spirit.

Sample code with fantasy API to "queue up" a visit in case one is currently running:

```js
function navigate(this: Swup, url, options) {
  /* Wait for current visit to finish before queuing new one */
  if (this.visit.running) {
    this.hooks.once('visit:end', () => this.navigate(...args));
    return;
  }
  /* Handle visit normally */
}
```

daun:
Brain dump over 🕳️

hirasso:
Great stuff @daun! The more I think about it, the more I feel like you are right and we should wait for the visit to finish and queue a subsequent visit, instead of continuously (always have to google how to spell this 😅) trying to cancel the current visit. I actually like your proposed snippet! We could even use the same mechanic as we did with `cancelled`, with an initially undefined `ended` property that could be marked as `@private`:

```js
function navigate(this: Swup, url, options) {
  /* Wait for current visit to finish before queuing new one */
  if (!this.visit.ended) {
    this.hooks.once('visit:end', () => this.navigate(...args));
    return;
  }
  /* Handle visit normally */
}
```
I only see one problem we still would need to solve: How do we handle *multiple* requests to `navigate` during a visit? The above snippet would queue them all and they would all be fired as soon as the current visit ends. I'm sure we'll be able to find an elegant solution for this one, too. Something like this maybe, outside of the `Visit` object:

```
function navigate(this: Swup, url, options) {
  /* Wait for current visit to finish before queuing new one */
  if (!this.visit.ended) {
    this.onVisitEnd = () => this.navigate(...args);
    return;
  }
  /* Handle visit normally */
}
```

...and then call that function each time a visit ends. (Default would be `() => {}`. That way we would ensure that

1. the last invocation of `navigate`
2. the `...args` are being preserved

hirasso:
I'll create a PR for the behavior described above so that we can discuss it.

hirasso:
I just realized we already have a way to tell if swup is currently navigating! `swup.navigating` 😄

hirasso:
Findings from trying to implement the queue functionality: 

- We'll likely need to start passing around the current visit object instead of relying on a class global.

hirasso:
Closing in favor of #836

PR #830, opened 2023-11-17
feat: allow safely working with dom events in TypeScript

hirasso:
**Description**

Exports a type `HookEvent`, so that `swup:...` events can be consumed more safely:

```js
import type { HookEvent } from "swup";

window.addEventListener("swup:all", (event) => {
  const hookName = (event as HookEvent).detail.hook;
  // @ts-expect-error
  const missing = (event as HookEvent).detail.missing;
});

```

**Alternatives considered**

Augmenting the global `GlobalEventHandlersEventMap`. Seems to be not be working with dynamic keys. Casting the event is the next best thing I could come up with :)

**Checks**

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -22,7 +22,7 @@ import type {
 	HookDefaultHandler,
 	HookOptions,
 	HookUnregister,
-	HookDOMEvent
+	HookEvent
 } from './modules/Hooks.js';
 import type { Plugin } from './modules/plugins.js';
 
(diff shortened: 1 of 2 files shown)


daun:
How about just `HookEvent`?

hirasso:
Yes, just changed it, our minds seem to be synchronised ;)

daun (approved):
🚀

PR #834, opened 2023-11-20
feat: queue navigation

hirasso:
Related: #825 

**Description**

- queue the next visit during running visits
- Pass around the visit object instead of relying on the global `swup.visit`

**TODO**

- [x] Convert the `Visit` object to a class
- [x] Fix all errors that prevent functional tests from passing
- [x] Handle `hooks.call` and `hooks.callSync` with missing `visit` arg

**Checks**

### DIFF
--- a/src/modules/Hooks.ts
+++ b/src/modules/Hooks.ts
@@ -2,7 +2,7 @@ import type { DelegateEvent } from 'delegate-it';
 
 import type Swup from '../Swup.js';
 import { isPromise, runAsPromise } from '../utils.js';
-import type { Visit } from './Visit.js';
+import { Visit } from './Visit.js';
 import type { FetchOptions, PageData } from './fetchPage.js';
 
 export interface HookDefinitions {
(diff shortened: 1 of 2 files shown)


daun:
Note: I have a feeling the below code contains a bug. The check for `!url || url === from` probably also triggers when going to the homepage, as that should be an empty string here. The second part should suffice, as we're already sanitizing everything inside `Location`. Need to check, though.

```ts
this.hooks.callSync('link:click', visit, { el, event }, () => {
  const from = visit.from.url ?? '';
  event.preventDefault();
  if (!url || url === from) {
    /* ... */
  }
});
```

daun:
Tests are passing again 🎉 The legacy argument order is now also handled correctly.

daun:
>   if (!url || url === from)

To follow up on this after some research: `URL.pathname` is never empty and will always have a leading `/` in front of it, so the comparison here is safe and doesn't need addressing :)

hirasso:
Woah, I actually also was surprised by this. PR towards the MDN article? :)

PR #836, opened 2023-11-20
Manage visits via state

daun:
**Description**

- Combine/extend @hirasso's 2 PRs: [rapid navigation](https://github.com/swup/swup/pull/825) + [queue navigation](https://github.com/swup/swup/pull/834)
- Add `Visit.state` property and `visit:abort` hook

**Features**

- Convert the `Visit` object to a class
- Pass around the `Visit` object to avoid global mutations
- Add a `state` property to the `Visit` object
  - created → queued → started → leaving → loaded → entering → completed / aborted / failed
- Queue new visit if the current visit is already past the `loaded` state
  - Otherwise, abort the visit and return early in strategic places
  - When aborting only during the `entering` phase, we can have fewer return clauses throughout the code
- Only run hooks if visit is not aborted
- Create `visit:abort` hook to allow undoing dom changes like in `visit:end`
- Always clear `is-leaving` class

**Notes**

- Regarding the naming (`visit.aborted`, `visit:abort`), I asked ChatGPT for a few suggestions and that's what it came up with. I switched to `abort` to see how it feels. I actually quite like it, slightly more than `cancel`. But we can switch back to `cancel` no problem.

**Checks**

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -9,7 +9,8 @@ module.exports = {
 	/* First exclude everything, then re-include /src, then exclude tests */
 	ignorePatterns: ['/*', '!/src', '*.test.ts'],
 	rules: {
-		'no-unused-vars': ['error', { vars: 'all', args: 'none' }],
+		'no-unused-vars': 'off',
+		'@typescript-eslint/no-unused-vars': ['error', { vars: 'all', args: 'none' }],
 		'@typescript-eslint/ban-ts-comment': ['error', { 'ts-ignore': 'allow-with-description' }],
 		'@typescript-eslint/unbound-method': 'off',
 		'@typescript-eslint/no-floating-promises': 'off',
--- a/src/Swup.ts
+++ b/src/Swup.ts
@@ -7,7 +7,8 @@ import { type DelegateEventUnsubscribe } from './helpers/delegateEvent.js';
 
 import { Cache } from './modules/Cache.js';
 import { Classes } from './modules/Classes.js';
-import { type Visit, createVisit } from './modules/Visit.js';
+import { createVisit } from './modules/Visit.js';
+import type { Visit } from './modules/Visit.js';
 import { Hooks } from './modules/Hooks.js';
 import { getAnchorElement } from './modules/getAnchorElement.js';
 import { awaitAnimations } from './modules/awaitAnimations.js';
@@ -202,7 +203,7 @@ export default class Swup {
 		await nextTick();
 
 		// Trigger enable hook
-		await this.hooks.call('enable', undefined, () => {
+		await this.hooks.call('enable', undefined, undefined, () => {
 			// Add swup-enabled class to html tag
 			document.documentElement.classList.add('swup-enabled');
 		});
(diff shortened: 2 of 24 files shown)


hirasso (approved):

hirasso:
Great news! Whats the next steps then? Ship it? 🤠

daun:
Just tested parallel animations, they're working great out of the box. Since we always skip the out animation on those, we'll always queue the next navigation, which is amazing.

daun:
Fragments seem very broken at the moment, not sure what the issue is here, but debugging at full throttle 🏄‍♂️

daun:
Fragments are actually perfectly fine. The problem is with more than two consecutive visits that are aborted. Queueing them works great, but aborting them creates flickering.

daun:
The only thing still broken are 3+ consecutive visits aborted during the out phase. See the third animation in the video below ("Abort fragment visits"). This has nothing to do with fragments, just that's the only test that visits 3+ pages.

I'm currently experimenting with adding more early returns, but it might be an issue with the css classes as well?

https://github.com/swup/swup/assets/22225348/4cc2fc97-6e8c-4061-b2f4-02ae35593ae6

daun:
Fixed it! There was one return statement missing before marking the visit as done. Now everything's working.

daun:
I've added another getter to the visit object: `visit.done` which is true for either `completed`, `aborted` or `failed`. Switched to using that as the official way of checking if we should handle or drop a visit. When testing with server errors and other exceptions, this was required to get it all working.

daun:
Ready for review 🥯

daun:
I'd suggest we merge this now but wait with the release until next week. I'd love to try and put both this and the animation controllers into the same release. If I've nothing to show by next week, let's drop this standalone as 4.5.

hirasso:
This all sounds AMAZING! 😻 Yes, let's wait till next week.

daun:
The only other question before merging this would be: how do we deal with hooks for aborted visits that only run once?

As currently implemented, the handlers will only be unregistered if they're actually executed. Meaning that, if a visit was aborted, and a handler is configured to run once, then it will trigger on the next visit that isn't aborted. This will garuantee that we're not creating situations where such a hook handler will never execute if the first occasion happens to be for an aborted visit.

I think that's how I'd expect things to work, when not explained otherwise.

hirasso:
> The only other question before merging this would be: how do we deal with hooks for aborted visits that only run once?

> As currently implemented, the handlers will only be unregistered if they're actually executed. Meaning that, if a visit was aborted, and a handler is configured to run once, then it will trigger on the next visit that isn't aborted. This will garuantee that we're not creating situations where such a hook handler will never execute if the first occasion happens to be for an aborted visit.

> I think that's how I'd expect things to work, when not explained otherwise. We should establish it as good practice to check if the visit matches the current one or was aborted, and only conditionally execute whatever's in the handler. In that case, it shouldn't be too much of an issue.

I'd also say that is what we want. A `once` handler for `visit:end` for example would be fired for the last visit that makes it through after aborting a few others.

hirasso (approved):
🧨

hirasso:
Don't forget to change the target to the `development` branch before merging :)

daun:
💥

PR #837, opened 2023-11-21
Native ViewTransition support

daun:
**Description**

- New `native` option to enable [View Transition](https://developer.mozilla.org/en-US/docs/Web/API/View_Transitions_API) support
- Works great in local playground for standard scenarios

**History**

- Supersedes #802
- Rebased on the recent updates from #836

**To-do**

- [x] Allow combining native + non-native transitions
- [x] Fallback strategy for unsupported browsers
- [x] Strategy for dynamic containers and fragments
- [x] Add functional tests

**Questions**

- How do we deal with dynamic containers and fragments? How can we mark them as "changing"?
  - ~~Use a new class `is-updating` to mark containers~~
  - We already have the `is-changing` class that works perfectly here

**Follow-ups**

- We should think about the concept of pluggable/modular animation handlers (Native, CSS, JS, None)
- Something like that would reduce bundle size and improve control flow

**Checks**


hirasso:
In my tests, `renderPage` is still adding classes to the HTML / the containers if `native: true`. Checking for `this.options.native` before adding the classes is working fine for me: 
```js
if (visit.animation.animate && !this.options.native) {
	this.classes.add('is-rendering');
}
// later...
if (visit.animation.animate && !this.options.native) {
	// Make sure to add these classes to new containers as well
	this.classes.add('is-animating', 'is-changing', 'is-rendering');
	if (visit.animation.name) {
		this.classes.add(`to-${classify(visit.animation.name)}`);
	}
}
```
Should I push this here?

daun:
Thinking more about this, the perfect way to go here is allowing native and non-native transitions at the same time. This solves two problems at once: fallback animations if native view transitions aren't supported, as well as combined animations where it makes sense. 

The only way this can work is to run out-animations first, run native view transitions after, and run in-transitions last. There's no way of running them in parallel as native view transitions are basically just static screenshots moving across the screen.

daun:
An example that's working great locally., with native view transitions + fallbacks using keyframes.

```css
.page-transition {
  view-transition-name: page;
}

::view-transition-old(page) {
  animation: 0.4s ease-in-out both fade-out;
}

::view-transition-new(page) {
  animation: 0.4s ease-in-out both fade-in;
}

/* Fallback for non-supported browsers */
.swup-not-native.is-changing .page-transition {
  animation: 0.4s ease-in-out both move-out;
}

.swup-not-native.is-changing.is-rendering .page-transition {
  animation: 0.4s ease-in-out both move-in;
}

@keyframes fade-out {
  from { opacity: 1; }
  to { opacity: 0; }
}

@keyframes fade-in {
  from { opacity: 0; }
  to { opacity: 1; }
}
```

daun:
The example above has one flaw: the native transition is executed in parallel as a cross-fade, whereas the non-native transition is executed in sequence. The easiest way to make both behave the same is adjusting the keyframes to always occur in parallel by adding a delay:

```css
::view-transition-old(page) {
  animation: 0.3s ease-in-out both move-out;
}

::view-transition-new(page) {
  animation: 0.3s 0.3s ease-in-out both move-in;
}
```

The more complictaed way is not using native transitions and using the Parallel Plugin to handle this.

hirasso:
> The example above has one flaw: the native transition is executed in parallel as a cross-fade, whereas the non-native transition is executed in sequence. The way to make both behave the same is adjusting the keyframes to always occur in parallel by adding a delay:

You mean the other way around, right? 🙃 "The way to make both behave the same is adjusting the **view transitions** to also occur **in serial** by adding a delay:"

...except if parallel plugin is active, right? 🤪🤯

hirasso:
...I have a slightly bad gut feeling about allowing both class-based animations and native view transitions together...

daun:
> You mean the other way around, right? 🙃

Ah, yes, of course!

Parallel Plugin shouldn't be a problem here, though. It uses different class names than normal animations. That's probably a question of documentation but I'd argue that a move to native transitions will make Parallel Plugin 100% redundant anyway.

daun:
@hirasso I added back the `native` option. Turns out it's strictly necessary: just running `document.startViewTransition` for every visit will insert the default 300ms `root` crossfade between all animations, so we need to make sure people enable it themselves and add any appropriate fallbacks. This will be a question of good documentation.

daun:
@hirasso Tests are passing, new tests for native transitions added. Ready for review :)

hirasso (approved):
If it works, it works! 🤠🤠

PR #845, opened 2023-12-08
Standardize usage of globals

daun:
**Description**

- Standardize the usage of global properties on the window: `window.location` and `window.history`
- Avoid ambiguity and confusion with local variables
- Use `location.assign` instead of `location.href = *` for (I think) better semantics
- Add test for expected behavior on errors: remove current history entry, force-reload page

**Checks**

### DIFF
--- a/src/Swup.ts
+++ b/src/Swup.ts
@@ -242,7 +242,7 @@ export default class Swup {
 		const { origin, url, hash } = Location.fromUrl(href);
 
 		// Ignore if the new origin doesn't match the current one
-		if (origin !== window.location.origin) {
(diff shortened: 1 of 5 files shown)


hirasso (approved):
Nice! Didn't even know `window.location.assign()` was a thing :)

daun:
> Nice! Didn't even know `window.location.assign()` was a thing :)

Yeah, me neither before this! I've stumbled up on it while looking for a more elegant solution to the error redirects by changing it to `location.replace`, but it turns out that Georgy was way ahead again with the current solution 🤠

Issue #848, opened 2023-12-09
Ignore errors in user hooks


daun:
### Describe the problem 🧐

- Currently, exceptions in any hook handlers will always trigger a native `location.assign` as failure strategy
- This makes sense for errors in the default hook handlers as they contain the main business logic
- User hooks, however, usually don‘t break internal logic if there‘s a type error, division by zero, etc.
- This makes debugging hard as the native page load will clear the console and hide the actual error

### Describe the propsed solution 😎

- Catch errors in user hook handlers and log to the console
- Don‘t catch errors in the default hook handlers
- This should keep visits running even with errors in user hook handlers

### Alternatives considered 🤔

- Try/catch inside user hook handlers
- A bit cumbersome?

### How important is this feature to you? 🧭

Nice to have

### Checked all these? 📚

- [X] I have read and searched the [official docs](https://swup.js.org)
- [X] I have checked [discussions](https://github.com/orgs/swup/discussions) and [existing issues](https://github.com/issues?q=is:issue+org:swup) for related problems
- [X] I agree to follow this project's [Code of Conduct](https://github.com/swup/.github/blob/master/CODE_OF_CONDUCT.md)
- [X] I have provided all necessary information to the best of my knowledge

daun:
Implemented in #849

PR #849, opened 2023-12-10
Ignore errors in user hooks

daun:
**Description**

- Attempt to tackle #848 
- Make swup more forgiving when it comes to errors in user code
- Catch and log exceptions in user hooks to keep visits running
- Only throw and force-reload for exceptions in internal default handler

**Scenario**

- Mistyped a variable name in a user hook → swup now always reloads, making this hard to figure out in the console

**Checks**

### DIFF
--- a/src/modules/Hooks.ts
+++ b/src/modules/Hooks.ts
@@ -354,7 +354,7 @@ export class Hooks {
 
 		const { before, handler, after } = this.getHandlers(hook, defaultHandler);
 		await this.run(before, visit, args);
-		const [result] = await this.run(handler, visit, args);
+		const [result] = await this.run(handler, visit, args, true);
 		await this.run(after, visit, args);
 		this.dispatchDomEvent(hook, visit, args);
 		return result;
@@ -383,7 +383,7 @@ export class Hooks {
 		const [visit, args, defaultHandler] = this.parseCallArgs(hook, arg1, arg2, arg3);
 		const { before, handler, after } = this.getHandlers(hook, defaultHandler);
 		this.runSync(before, visit, args);
-		const [result] = this.runSync(handler, visit, args);
+		const [result] = this.runSync(handler, visit, args, true);
 		this.runSync(after, visit, args);
 		this.dispatchDomEvent(hook, visit, args);
 		return result;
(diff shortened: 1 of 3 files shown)


src/modules/Hooks.ts
```diff
@@ -416,25 +416,30 @@ export class Hooks {
...
 	// Overload:  running user handler: expect no specific type
 	protected async run<T extends HookName>(registrations: HookRegistration<T>[], visit: Visit | undefined, args: HookArguments<T>): Promise<unknown[]>; // prettier-ignore
 	// Implementation
 	protected async run<T extends HookName, R extends HookRegistration<T>[]>(
 		registrations: R,
-		visit: Visit | undefined,
-		args: HookArguments<T>
+		visit: Visit | undefined = this.swup.visit,
+		args: HookArguments<T>,
+		rethrow: boolean = false
```
hirasso, line 427: Wouldn't `throw` be descriptive enough? I think I'd prefer that. Or is naming it `rethrow` following some kind of convention for that kind of behavior?

daun: I had throw, but it's a reserved word in JS :(

hirasso: ah! Makes sense 😁

hirasso: How about `throwErrors` then? I know you prefer one-word variables, but that I would understand immediately

daun: How do you feel about `throws`? with an s?

hirasso: I don't like it better than rethrow.

hirasso: Did you ask ChatGpt? 😉

hirasso: Alternatively, maybe we could call the variable `internal` or `builtin`. And then throw errors if `internal` is true.

daun: I did, and it said `rethrow` is the common word for that behavior. [Google seems to agree rethrowing is a thing](https://www.google.com/search?q=rethrow+error&oq=rethrow+error).

daun: I frankly don't mind much, it's an internal helper function with no public API, so let's go with `throwError` if `rethrow` doesn't make sense for you.

hirasso (approved):

src/modules/Hooks.ts
hirasso: Thanks for googling that for me 🤦😅 ...lets go with rethrow.

Issue #847, opened 2023-12-09
Feature: Allow aborting running visits


daun:
### Describe the problem 🧐

- Continuation of #844
- In browsers, hitting the `Esc` button stops the page from loading
- Swup currently doesn't have a way of specifically cancelling a running visit

### Describe the propsed solution 😎

Extend the functionality of `visit.abort()`:

- Cancel the running `fetch` request
- Revert the url bar change
- Undo the out-animation to restore the content

There are two scenarios of aborting visits:

- **A — Abort a visit in favor of a new one**: already supported, much simpler
- **B — Abort a visit and revert to previous state**: new feature, a bit more complex

Questions:

- Is "abort" the right terminology for both? Can both be supported with the same hook? One abort is "final" and stops everything, one abort is "intermediate" and merely transitions into the next visit
- We could give the `visit:abort` hook an argument, `{ final: true }` that indicates this. Feels a bit strange.
- Example: hiding the progress bar on `visit:abort` only makes sense for scenario B, when the abort is "final".
- Do other libraries support this? Turbo, Pjax, Livewire, etc.

### Alternatives considered 🤔

- Not implement the feature at all. It will probably add considerable complexity while not being a very common use case.
- Extend existing or create new hooks to allow implementing this in userland
- Just canceling the underlying fetch request is probably not enough as it won't fire `end` or `abort` hooks

### How important is this feature to you? 🧭

Nice to have

### Notes

- I think we're getting into state machine territory here
- From a cursory look over how the plugins determine the start and end of a visit, how the `swup.navigating` is set/unset, how thrown errors return early without 'finishing' a visit, this seems rather complex
- Started a [draft PR](https://github.com/swup/swup/pull/850) to explore the complexity and scope of this

daun:
Closing as not planned.

PR #850, opened 2023-12-10
Allow aborting and undoing visits

daun:
**Description**

- Allow aborting running visits directly, without a queued next visit: `visit.abort()`
  - Implementing a global stop button would be as easy as `swup.visit.abort()`
- Allow undoing changes when aborting visits: `visit.abort(true)`
  - Remove animation classes
  - Reset browser url to previous url
- Only allows aborting visits that have not yet started entering
- Only allows aborting non-popstate visits
- Does not abort any running fetch requests; keeping track of that was a nightmare
- Undoing a visit triggers a new `visit:undo` hook
  - Useful for progress plugin and parallel plugin to clean up
- Adds 0.18kB to the bundle size

**API**

I've found the method signature of `visit.abort(undo: boolean = false)` kind of logical and self-explanatory. We can always upgrade the method signature in the future by allowing more options like `true | { undo: boolean; lorem: string }` but I don't think there will by any in the foreseeable future.

- `visit.abort()` for a soft abort ("ignore")
- `visit.abort(true)` for a hard abort ("undo") 
- Precedent: `window.location.reload(force: boolean = false)` for soft/hard reload

**Required changes**

- [x] Make `visit.abort()` public
- [x] Make `visit.abort()` trigger the logic for aborting the visit
- [x] Add `visit:undo` hook
- [x] Save visit id and history action (push/replace) into the history state to allow undoing url changes
- [x] Save previous page's hash into the visit object to allow fully restoring the previous url

**Questions**
- Is this something we want to support?
- What about JS animations? native animations? parallel animations? 😵‍💫
  - JS plugin could create a new function: `out`, `in`, and `undo`? Or just call `out`?
  - Parallel animations can't be aborted as they happen after the content is already loaded
  - People can hook into `visit:undo` to undo custom stuff

**Checks**

**Additional information**

Discussed in #847


daun:
@hirasso Happy to get on a call whenever convenient, to walk you through the changes and get your input.

hirasso:
Works nicely in my tests with a simple CSS transition! Out of curiosity I also tried to immediately undo a visit on `visit:start` which doesn't work (yet?). While certainly one could argue about sense and nonsense of this, I'd still find it good to support it. I was able to get it to work by applying the following changes in `navigate.ts`:

1. Move the history push/replace state logic and the `this.currentPageUrl` mutation before the `visit:start` hook
2. Check for `visit.done` directly after `visit:start` before continuing

The relevant code:

```js
// Create/update history record if this is not a popstate call or leads to the same URL
if (!visit.history.popstate) {
	// Add the hash directly from the trigger element
	const newUrl = visit.to.url + visit.to.hash;
	if (visit.history.action === 'replace' || visit.to.url === this.currentPageUrl) {
		updateHistoryRecord(newUrl, { visit: visit.id });
	} else {
		this.currentHistoryIndex++;
		createHistoryRecord(newUrl, { visit: visit.id, index: this.currentHistoryIndex });
	}
}

this.currentPageUrl = getCurrentUrl();

await this.hooks.call('visit:start', visit, undefined);

// Check if failed/aborted in the meantime
if (visit.done) return;

visit.state = VisitState.STARTED;
// ... the rest of `performNavigation`
```
The above changes unfortunately break the scrolling test [`appends the hash if changing visit.to.hash on the fly`](https://github.com/swup/swup/blob/c2107c67876d51acdf961909ce02dee4dff22db5/tests/functional/scrolling.spec.ts#L64), since the URL has already updated when `visit:start` runs. That's unfortunate...

I see two alternatives:

1. don't validate the visit's ID against the current history entry, just trust that it's the right one:
```diff
- // Aborting most recent visit? Undo history and url bar changes
- if (state.visit === visit.id) {
  this.currentPageUrl = visit.from.url + visit.from.hash;
  // ...
```
2. Leave it as is and mention it in the docs. The earliest hook where a visit can be undone is before `visit:transition`. I don't really see a use case for using it in `visit:start` anyways.

hirasso:
...we could maybe think about moving the history update block right after `visit:start`, to make it possible to undo a visit from a `page:load` hook...

src/modules/navigate.ts
```diff
@@ -232,3 +229,39 @@ export async function performNavigation(
...
+			if (state.action === 'replace') {
+				updateHistoryRecord(visit.from.url + visit.from.hash);
+			} else {
+				window.history.back();
+			}
+		});
+	}
+
+	// Only set this here so that hooks still get called
+	// Probably set in visit.abort() right after visit.undo(), but let's make sure
```
hirasso, line 265: Needs adjustment to updated architecture. `visit.abort(true)` instead of `visit.undo()`

daun:
Nice! Let's add a check for `visit.done` after the `visit:start` hook, we definitely should have that anyway.

You're right, visits need to be abortable from `visit:start`. What do you mean by "doesn't work yet"? In my head, the previous suggested check for `visit.done` right after `visit:start` should be sufficient for that at there's nothing yet to undo at this point. But I'm probably missing some part here...

hirasso:
@daun what was the scenario you were thinking of when implementing the safeguard with the check for the visit‘s id in the history state?

daun:
@hirasso That's to make sure we're only undoing a visit if this visit is the current (= last) history entry.

daun:
@hirasso I've added your suggestions and added a currently failing test. The test failures are very exotic, probably related to an unwanted exception and/or reload. Needs more work :)

hirasso:
I'd really love to move the history updating logic to the very start, just before `visit:start`. That would ensure that the URL during `visit:start` would be consistently equal to `visit.to.url`, no matter if it's a normal or a history visit. I'll create an issue for that.

daun:
I just noticed an error after our previous discussion about the url bar state during popstate visits. When creating a popstate visit, the previous hash in the visit object is wrongly taken from the current url -- which is the new one in popstate visits. We currently don't track the current hash, only the current url without hash. Just leaving this here so I don't forget about it.

hirasso:
Another error with the current implementation: setting the hash on the fly won't work with history navigation

daun:
@hirasso True! That might be related to the previous point as we're not storing the current hash anywhere.

daun:
@hirasso I've rebased this onto the current master and made sure the tests are working as expected.

What's our current position on this? Do we want this as a feature? There's a few things I like here — front of all, using `swup.location` with an actual `Location` object in stead of `swup.currentPageUrl` as a string, to be able to get the current hash as well. But we might just implement that without the part about aborting and undoing visits.

daun:
@hirasso Just wondering what you meant by "setting the hash on the fly won't work with history navigation" — what's your usage scenario and what's missing to get this working?

hirasso:
@daun I think its better to explain what I mean in a call

hirasso:
Regarding aborting visits: I'm leaning towards not implementing this, after the hickups we recently already had with `visit.done`

daun:
@hirasso I'm also in favor of dropping this for now. It was a nice challenge to think through and I think we landed on something solid, but the added complexity is for a feature I'd honestly never find use for, which will make supporting this bothersome.

daun:
Closing this for now. Let's keep the branch around and see if we ever feel like digging it up again 🕳️

PR #858, opened 2024-01-05
fix: transition classes mutation if pausing before `content:replace`

hirasso:
Fixes #857 

**The Problem**

When pausing before `content:replace`, CSS transitions can get out of sync with the actual content replacement, as described in #857. The option [`awaitAssets`](https://swup.js.org/plugins/head-plugin/#awaitassets) in Head Plugin, for example, will wait for all new assets to be loaded before proceeding with `content:replace`. 

The cause for this is that the classList mutation of `is-leaving` and `is-rendering` is currently being done _outside_ of `content:replace`, while in fact it should happen _right before_ the actual replacement, as described in the docs:

> `is-leaving`: [...] Removed **right before the content is replaced**. 
> `is-rendering`: Added **right before the content is replaced** [...]

**The Solution**

Move the classList mutations inside the `content:replace` hook. 

**Questions**

Could this change have unforeseen consequences? I can't think of any, tests are passing. I'd say we should definitely fix this, as otherwise `awaitAssets` is broken right now for certain setups.

**Code to reproduce the bug with the current version of swup**

The following code will reliably:

- animate an overlay for every visit
- replace the content after a **2 second delay** after the out animation started (bad)

```html
<body>
  <div class="transition-overlay"></div>
  <div id="swup"><!-- ... --></div>
</body>
```

```css
.transition-overlay {
  position: fixed;
  top: 0;
  left: 0;
  right: 0;
  bottom: 0;
  z-index: 999;
  background-color: rgb(64, 84, 145);
  opacity: 0;
  transform: translate3d(0, -100%, 0);
  pointer-events: none;
}

html.is-changing .transition-overlay {
  transition: transform 400ms ease-in-out;
  opacity: 1;
}

html.is-animating .transition-overlay {
  transform: translate3d(0, 0, 0);
}

html.is-rendering .transition-overlay {
  transform: translate3d(0, 100%, 0);
}
```

```js
const swup = new Swup();
swup.hooks.before("content:replace", async (visit: Visit) => {
  await sleep(2000); // wait for 2 seconds before proceeding
});

function sleep(timeout = 0): Promise<void> {
  return new Promise((resolve) => setTimeout(() => resolve(undefined), timeout))
}
```

**Checks**

- [x] Tested manually in local playground
- [x] New tests added

### DIFF
--- a/src/modules/renderPage.ts
+++ b/src/modules/renderPage.ts
@@ -14,8 +14,6 @@ export const renderPage = async function (this: Swup, visit: Visit, page: PageDa
 
 	const { url } = page;
 
-	this.classes.remove('is-leaving');
-
 	// update state if the url was redirected
 	if (!this.isSameResolvedUrl(getCurrentUrl(), url)) {
 		updateHistoryRecord(url);
@@ -23,13 +21,13 @@ export const renderPage = async function (this: Swup, visit: Visit, page: PageDa
 		visit.to.url = this.currentPageUrl;
 	}
 
-	// only add for animated page loads
-	if (visit.animation.animate) {
-		this.classes.add('is-rendering');
-	}
-
 	// replace content: allow handlers and plugins to overwrite paga data and containers
 	await this.hooks.call('content:replace', visit, { page }, (visit, { page }) => {
+		this.classes.remove('is-leaving');
+		// only add for animated page loads
+		if (visit.animation.animate) {
+			this.classes.add('is-rendering');
+		}
 		const success = this.replaceContent(page, { containers: visit.containers });
 		if (!success) {
 			throw new Error('[swup] Container mismatch, aborting');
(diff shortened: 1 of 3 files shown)


daun (approved):
Great find! Let's release this as a patch release; I don't think it will break anything, but if it does, we'll fix that as well :)

daun:
The issue probably never came up during testing because we usually test with animations where leaving and entering are mirrored and we haven't needed `is-leaving` or `is-rendering`.

daun:
I'm not a fan of the fact that we're adding css-specific classes inside a hook for replacing content, but that's already been the case since we added local container scope. I'd love to explore ways of separating these, but that's for another time 🤠

daun:
Just realized we'll also be releasing the previously merged [PR](https://github.com/swup/swup/pull/849) for gracefully handling errors in user hook handlers. I'd say it's still a patch release since we're only correcting behavior that wasn't originally intended/conceived, but what do you think?

hirasso:
I agree :)

PR #868, opened 2024-01-26
Fix reflow helper being optimized away

daun:
**Description**

- The `forceReflow` helper seems to be getting optimized away by minifiers
- Accessing `el.offsetHeight` forces the reflow, but [to minifiers it looks like a side-effect free getter](https://github.com/terser/terser/issues/464)
- Switching to `getBoundingClientRect()` fixes this as [it's a method and methods have side-effects](https://github.com/angular/angular-cli/issues/17494#issuecomment-635592036)
- Required to fix this [issue in parallel-plugin](https://github.com/swup/parallel-plugin/issues/13) where the helper is removed in UMD builds
- Tested to work in local playground

**Checks**

### DIFF
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -46,12 +46,11 @@ export function runAsPromise(func: Function, args: unknown[] = []): Promise<unkn
 
 /**
  * Force a layout reflow, e.g. after adding classnames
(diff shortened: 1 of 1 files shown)


hirasso (approved):
Very interesting!

PR #869, opened 2024-01-26
Store new document in visit object

### LINKED ISSUE #862: Feature: Store parsed html document in visit object
### Describe the problem 🧐

The swup.js library currently parses the entire HTML document internally using the DOMParser API. However, the parsed HTML document is not exposed or accessible to the users of the library. The library only returns the HTML document as a string.

Parsing an HTML document is a relatively slow process. In scenarios where the developer needs to work with the parsed HTML document, they have to parse the same HTML document again. This results in unnecessary duplication of work and performance overhead.

To improve the efficiency and flexibility of the library, I propose adding an option that allows users to control whether the parsed HTML document should be included in the visit object or not. When this option is enabled, the visit object should include a reference to the parsed HTML document.

I believe this feature would be a valuable addition to the swup.js library and would greatly improve its usability and performance in certain scenarios.

Thank you for considering this feature request.

### Describe the propsed solution 😎

const swup = new Swup({
  includeParsedHtmlInVisit: true, // New option
  // Other options...
});

swup.hooks.on('content:replace', (visit) => {
  const parsedHtml = visit.to.html.parsed; // Parsed HTML document
  // ...
});

### Alternatives considered 🤔

- Parse the same document string again
- Don't parse it again - search in string using regex

### How important is this feature to you? 🧭

Nice to have

### Checked all these? 📚

- [X] I have read and searched the [official docs](https://swup.js.org)
- [X] I have checked [discussions](https://github.com/orgs/swup/discussions) and [existing issues](https://github.com/issues?q=is:issue+org:swup) for related problems
- [X] I agree to follow this project's [Code of Conduct](https://github.com/swup/.github/blob/master/CODE_OF_CONDUCT.md)
- [X] I have provided all necessary information to the best of my knowledge

daun:
**Description**

- Store the parsed `Document` of the new page in the visit object as `visit.to.document`
- Allow reuse in plugins and consumer hooks to improve performance
- Remove the parsed document when the visit ends, is aborted, or throws an exception
- Read the containers from the visit object directly instead of duplicating them as arguments

**Questions**

- Are we rewriting the html in any plugins where we'd also need to rewrite the document?

**Checks**

**Context**

- Closes #862

### DIFF
--- a/tests/unit/replaceContent.test.ts
+++ b/tests/unit/replaceContent.test.ts
@@ -1,6 +1,7 @@
-import { describe, expect, it, vi, afterEach } from 'vitest';
+import { describe, expect, it, vi, afterEach, beforeEach } from 'vitest';
 import Swup from '../../src/Swup.js';
 import type { PageData } from '../../src/modules/fetchPage.js';
+import { Visit, createVisit } from '../../src/modules/Visit.js';
 import { JSDOM } from 'jsdom';
 
 const getHtml = (body: string): string => {
@@ -12,19 +13,22 @@ const getHtml = (body: string): string => {
 	`;
 };
 
-const mockPage = (body: string): PageData => {
-	return {
-		url: '',
-		html: getHtml(body)
-	};
-};
-
 const stubGlobalDocument = (body: string): void => {
 	const dom = new JSDOM(getHtml(body));
 	vi.stubGlobal('document', dom.window.document);
 };
 
+class SwupWithPublicVisitMethods extends Swup {
+	public createVisit = createVisit;
+}
+
+const swup = new SwupWithPublicVisitMethods();
+let visit: Visit;
+
 describe('replaceContent', () => {
+	beforeEach(() => {
+		visit = swup.createVisit({ to: '' });
+	});
 	afterEach(() => {
 		vi.unstubAllGlobals();
 	});
@@ -36,12 +40,13 @@ describe('replaceContent', () => {
 			<div id="container-3" data-from="current"></div>
 		`);
 
-		const page = mockPage(/*html*/ `
+		const html = `
 			<div id="container-1" data-from="incoming"></div>
-			<div id="container-2" data-from="incoming"></div>`);
-		const swup = new Swup();
+			<div id="container-2" data-from="incoming"></div>`;
+		visit.to.document = new DOMParser().parseFromString(html, 'text/html')
 
-		const result = swup.replaceContent(page, { containers: ['#container-1', '#container-2'] });
+		const swup = new Swup();
+		const result = swup.replaceContent(visit, { containers: ['#container-1', '#container-2'] });
 
 		expect(result).toBe(true);
 		expect(document.querySelector('#container-1')?.getAttribute('data-from')).toBe('incoming');
@@ -54,13 +59,14 @@ describe('replaceContent', () => {
 			<div id="container-1" data-from="current"></div>
 		`);
 		const warn = vi.spyOn(console, 'warn');
-		const page = mockPage(/*html*/ `
+		const html = `
 			<div id="container-1" data-from="incoming"></div>
 			<div id="container-2" data-from="incoming"></div>
-			`);
+			`;
+		visit.to.document = new DOMParser().parseFromString(html, 'text/html')
 
 		const swup = new Swup();
-		const result = swup.replaceContent(page, { containers: ['#container-1', '#missing'] });
+		const result = swup.replaceContent(visit, { containers: ['#container-1', '#missing'] });
 
 		expect(result).toBe(false);
 		expect(warn).not.toBeCalledWith(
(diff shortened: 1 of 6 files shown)


hirasso:
> Are we rewriting the html in any plugins where we'd also need to rewrite the document?

I can't think of any

tests/unit/replaceContent.test.ts
```diff
@@ -54,13 +59,14 @@ describe('replaceContent', () => {
 			<div id="container-1" data-from="current"></div>
 		`);
 		const warn = vi.spyOn(console, 'warn');
-		const page = mockPage(/*html*/ `
+		const html = `
```
hirasso, line 62: The `/* html */` comment tells this vscode extension to correctly syntax-highlight the html template literal:

https://marketplace.visualstudio.com/items?itemName=Tobermory.es6-string-html

daun: Good to know! I've added them back.

daun:
@hirasso Feel free to give this a spin locally! Tests are passing and I've been toying with it in the playground, but it's best to get more eyes on this as we'll be probably start relying on the new property in plugins. At least for body class and head plugin, from what I can think of.

hirasso:
So, I gave this a spin. It's working well! Of course, this makes it easy to break swup:

```
swup.hooks.before("content:replace", (visit: Visit) => {
  visit.to.document?.querySelector('#swup')?.remove();
});
```

So it's probably worth to mention in the docs that people should be careful when working with `visit.to.document`.

Another thing I noticed while testing this: on `page:load`, `visit.to.html` and `visit.to.document` aren't accessible, as we are only putting them in the `visit` after that hook. I imagine people would intuitively reach for the `page:load` hook to grab the document as early as possible. Is there a reasoning behind this or can we move it inside the hook?

daun:
@hirasso Those are good points.

> Of course, this feature makes it easy to break swup

Breaking swup has become easier and easier with replaceable hooks and a writable visit object, but I see that as a net-positive. I like tools I can bend to work the way If want to 🤠 Mentioning that in the docs is probably a good idea, but I wouldn't lose sleep over it.

> people should be careful when working with visit.to.document and remember to cloneNode(true) things if they want to extract elements from it

Your comment reminded me to double-check we're not actually removing any elements from the new document while replacing them in `replaceContent`. I'll do that tomorrow.

> on page:load, visit.to.html and visit.to.document aren't accessible, as we are only putting them in the visit after that hook

I agree — having access to the new document would be handy and intuitive in `page:load`. There is a reason it's currently implemented outside of the hook: the hook is supposed to be replaceable without duplicating its logic, e.g. by the preload plugin that returns a preload promise instead. Adding this logic to the hook handler would require the preload plugin to duplicate this logic, returning us to the previous state of duplicating logic across the core and its plugins. Having this outside of a hook ensures it always happens.

Can we think of any other way of making this available? Personally, I wouldn't know what to do with that data since the timing of `page:load` is random and there's more useful hooks later during the lifecycle, but consumers would be kind of right to expect it there.

hirasso:
> Breaking swup has become easier and easier with replaceable hooks and a writable visit object, but I see that as a net-positive. I like tools I can bend to work the way If want to 🤠 Mentioning that in the docs is probably a good idea, but I wouldn't lose sleep over it.

I sure love the hackability of _modern_ swup 🥳 Lets just not forget to mention this in the docs. OR not mention it at all for now 🙃

> Your comment reminded me to double-check we're not actually removing any elements from the new document while replacing them in replaceContent. I'll do that tomorrow.

Nice!

> I agree — having access to the new document would be handy and intuitive in page:load. There is a reason it's currently implemented outside of the hook: the hook is supposed to be replaceable without duplicating its logic, e.g. by the preload plugin that returns a preload promise instead. Adding this logic to the hook handler would require the preload plugin to duplicate this logic, returning us to the previous state of duplicating logic across the core and its plugins. Having this outside of a hook ensures it always happens.

> Can we think of any other way of making this available? Personally, I wouldn't know what to do with that data since the timing of page:load is random and there's more useful hooks later during the lifecycle, but consumers would be kind of right to expect it there.

Let's leave it as-is. I was probably making up a problem where none exists 😄

daun:
@hirasso You had the correct hunch! The current implementation was still missing `cloneNode(true)` to avoid modifying the document and inadvertently removing the incoming containers. Would have totally missed that. Fixed, and added a unit test.

hirasso:
Nice! Shall I re-approve then?

hirasso:
Does parallel plugin overwrite `replaceContent`? I'm AFK at the moment and can't check

hirasso (approved):

daun:
Parallel Plugin only hooks before and after `content:replace`. It leaves the default handler in place so that it can handle any non-parallel containers still left in `visit.containers`.

hirasso:
Do we have to check other plugins as well for this issue?

PR #876, opened 2024-02-14
Remove compatibility checks that are no longer required

daun:
**Description**

- Remove some compatibility checks and fixes no longer required given our current browser support matrix
  - Any browsers that support our UMD build syntax support `Promise` and `CSS.escape`
  - Our current browser support after polyfilling is Chrome 66, so the bug in `toMs` is no longer an issue
- I've removed the (now empty) `checkRequirements` method, we can always add it back in the future
- Changes are non-breaking, as other modern browser features would be an issue long before these are
- Slims the bundle down by 2% or 0.13K

**Checks**

### DIFF
--- a/src/Swup.ts
+++ b/src/Swup.ts
@@ -156,15 +156,7 @@ export default class Swup {
 
 		this.currentHistoryIndex = (window.history.state as HistoryState)?.index ?? 1;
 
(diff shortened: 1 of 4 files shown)


hirasso (approved):
💅

PR #878, opened 2024-02-16
Add Browserstack note

daun:
- Browserstack [supports open-source projects](https://www.browserstack.com/open-source) with subsidized (= free) plans
- The only requirement is the text »This project is tested with BrowserStack.« in the readme
- I wasn't able to check if that plan includes accessibility testing (our main goal here), but it's worth a shot


hirasso (approved):
🤞

PR #901, opened 2024-06-06
 Restore focus to persisted element after replacing content

ovenum:
This PR restores focus to a DOM Element persisted with `data-swup-persist` if it was the focused element before content replacing took place.

This is useful for input elements that trigger a search and should remain focused when the result is added to the DOM.

To further illustrate what this PR tries to solve please see the example on replit [here](https://replit.com/@ove2/data-swup-persist-bug-report)

**Checks**

**Additional information**

- Tests are missing and would appreciate help in adding them if this PR should be merged. 
- Documentation for `data-swup-persist` should reflect that focus is persisted.

### DIFF
--- a/src/modules/replaceContent.ts
+++ b/src/modules/replaceContent.ts
@@ -11,6 +11,9 @@ export const replaceContent = function (this: Swup, visit: Visit): boolean {
 	const incomingDocument = visit.to.document;
 	if (!incomingDocument) return false;
 
+	// Store currently focused element
+	const activeElement = document.activeElement;
+
 	// Update browser title
 	const title = incomingDocument.querySelector('title')?.innerText || '';
 	document.title = title;
@@ -43,6 +46,11 @@ export const replaceContent = function (this: Swup, visit: Visit): boolean {
 		const replacement = query(`[data-swup-persist="${key}"]`);
 		if (replacement && replacement !== existing) {
 			replacement.replaceWith(existing);
+
+			// Restore focus to the persisted element if it was focused before
+			if (existing === activeElement) {
+				existing.focus();
+			}
 		}
 	});
 


daun:
@hirasso What's your opinion on this? With just the core in mind, this sounds like a good idea, no? Where it confuses me is the a11y stuff — we'd 1) lose focus 2) refocus the input 3) reset focus to the body. Might be confusing, but it also might just be debounced and totally fine. I'd suggest limiting this to inputs to keep this specific.

ovenum:
There could be another issue with persisted elements having client state that is only tracked in JS and input elements inside these elements. For example a persisted section nav-bar with filter menus and a search input.

If focus is only re-applied to input elements, it could be enough to also search the child elements of the persisted element for input elements.

How would this work with WebComponents that extend on Input elements?

hirasso:
Thank you @ovenum for bringing this up! 

I have recently been fighting with exactly this use case, as well.

I ended up not persisting my filter nav but rather managed as much state as possible on the server, and restored the focus using a11y-plugin (which provides custom  handling of [`visit.a11y.focus`](https://swup.js.org/plugins/a11y-plugin/#visit-a11y-focus)) and a custom hook:

```js
swup.hooks.before("content:replace", (visit) => {
    const focusID = document.activeElement?.getAttribute("data-focus-id");
    if (focusID) {
      visit.a11y.focus = `[data-focus-id="${focusID}"]`;
    }
  });
```
That way, I can control from my markup if the focus on an element should be restored when navigating.

While thinking about focus, we have to keep in mind, that the expected behavior for screen reader users is that the focus is _always_ set to the `<body>` for each new page visit. That's also what we did in [version 5 of @swup/a11-plugin](https://github.com/swup/a11y-plugin/releases/tag/5.0.0).

If we decide to proceed with this, It could be helpful to revisit what Astro decided to do with focus in their [`transition:persist`](https://docs.astro.build/en/guides/view-transitions/#maintaining-state) directive. 

Regarding the (Alpine.js?) state: usually, morphing and persisting is exactly meant for persisting state. (see [Alpine.js' Morph Plugin](https://alpinejs.dev/plugins/morph) or the headline in the Astro docs ("Maintaining State").

I think this is a hard problem that I'd love to discuss in a call between the three of us sometime soon, where we can look at our real-world use cases together and discuss if this can be generalized somehow without silently breaking anyone's implementation. 

It might turn out to be the cleanest solution to introduce a new attribute `data-swup-persist-focus` or the like, to give users the power (and responsibility!) of managing this without a custom hook as I used above. But then again, what should win if a11y-plugin is installed?! 😵‍💫😄 ...as I said, we should talk this through.

hirasso:
As discussed, we are closing this for now. Users can easily implement their own solution using a hook as described above. We should keep in mind to think about merging swup/a11y-plugin into core, to make accessibility a first class citizen in swup's ecosystem.

PR #904, opened 2024-06-09
Location getter + abortable check

daun:
**Description**

- Re-implement some changes from the `visit:abort` branch that make sense

**Location object**
- Store the current location as a `Location` object at `swup.location`
- Allows checking the current hash in addition to just the path/query
- Save the hash of the previous page in the visit object
- Deprecate `swup.currentPageUrl` in favor of `swup.location.url`

```js
if (swup.location.url === '/about') {}
if (swup.location.hash) {}
if (visit.from.hash === '#detail') {}
```

**Location update**
- Update browser url at earliest possible time

**Aborting visits**
- Add `visit.abortable` getter that defines when a visit can be aborted
- Visits that have started entering cannot be aborted
- ~~History visits cannot be aborted~~ (wrong assumption, needs to be abortable)
- Refactor aborting visits from `swup.navigate()`

**Note to self**
- Not allowing aborting history doesn't make too much sense here (we only abort visits if there's a new one)

**Checks**

### DIFF
--- a/src/modules/navigate.ts
+++ b/src/modules/navigate.ts
@@ -123,6 +123,18 @@ export async function performNavigation(
 
 		visit.state = VisitState.STARTED;
 
+		// Create/update history record if this is not a popstate call or leads to the same URL
+		const newUrl = visit.to.url + visit.to.hash;
+		if (!visit.history.popstate) {
+			if (visit.history.action === 'replace' || visit.to.url === this.location.url) {
+				updateHistoryRecord(newUrl);
+			} else {
+				this.currentHistoryIndex++;
+				createHistoryRecord(newUrl, { index: this.currentHistoryIndex });
+			}
+		}
+		this.location = Location.fromUrl(newUrl);
+
 		// Begin loading page
 		const page = this.hooks.call('page:load', visit, { options }, async (visit, args) => {
 			// Read from cache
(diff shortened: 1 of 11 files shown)


hirasso:
> Not allowing aborting history doesn't make too much sense here (we only abort visits if there's a new one)

It should read "allowing", not "Not allowing", right? :)

daun:
@hirasso No, I think that's what I actually meant to write. In the following sense: by not allowing history visits to be aborted, we now make it impossible to supersede history visits. That would be the case when clicking the back button a few times in quick succession. If the cache is disabled or history visits are animated, every one of these history visits will have to finish completely before queueing the next one. Hence, we should allow aborting history visits if there's a new visit. I need to remove that part from this PR.

The really not allowing to abort them was only necessary in the previous PR where one could abort a visit from anywhere at any point. But even there we should have added a special condition for quick history visits in succession.

This is all very confusing :/ I really wish we had a nice state machine 🤖

hirasso:
It's indeed pretty confusing... thank you for clarifying what exactly you meant.

daun:
Addressed the open points and fixed the tests. Ready for review :)

src/modules/navigate.ts
```diff
@@ -123,8 +117,21 @@ export async function performNavigation(
 
 	try {
 		await this.hooks.call('visit:start', visit, undefined);
+
 		visit.state = VisitState.STARTED;
 
+		// Create/update history record if this is not a popstate call or leads to the same URL
```
hirasso, line 123: Moving the bit where the URL updates feels pretty scary to me. I can recall various situations where we discovered quite late in the process that the exact moment where the URL was updated had a lot of implications.

Why did you move it?

daun:
This is doing a bit too much — slimmed down PR incoming :)

PR #912, opened 2024-06-19
Location getter

daun:
Version of #904 without the extra stuff...

**Description**

- Store the current location as a `Location` object at `swup.location`
- Save the hash of the previous page in the visit object
- Deprecate `swup.currentPageUrl` in favor of `swup.location.url`

```js
if (swup.location.url === '/about') {}
if (swup.location.hash) {}
if (visit.from.hash === '#detail') {}
```

**Checks**

### DIFF
--- a/src/Swup.ts
+++ b/src/Swup.ts
@@ -93,8 +93,12 @@ export default class Swup {
 	readonly hooks: Hooks;
 	/** Animation class manager */
 	readonly classes: Classes;
-	/** URL of the currently visible page */
-	currentPageUrl: string = getCurrentUrl();
+	/** Location of the currently visible page */
+	location: Location = Location.fromUrl(window.location.href);
+	/** URL of the currently visible page @deprecated Use swup.location.url instead */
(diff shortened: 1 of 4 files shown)


hirasso (commented):
Fine with me :) ...would you like to expand on the use case you have in mind for this?

daun:
Sure! I'm currently using this from a private fork with the following use case:

- The page transition depends on whether I'm coming from or going to a shopping cart modal
- The shopping cart modal is implemented client-side via url state, i.e. an anchor in the url. A url with `#cart` means the cart is open, a url without the anchor means it's closed
- The cart is a third-party service implementation I have no control over
- Checking the previous url currently works using `visit.from.url`, but it doesn't include the hash
- Adding the hash to the visit object is much easier if we track the current url from a `Location` object instead of a string, otherwise we'd need `swup.currentPageUrl` and `swup.currentPageHash` to store both

hirasso:
Nice! Thank you

hirasso (approved):

PR #933, opened 2024-07-23
Register hooks from constructor options

daun:
**Description**

- Add new `hooks` options for registering hook handlers when creating swup
- Accept hook modifiers à la Alpine/Vue: `content:replace.before`

**Example**

```js
const hooks = {
  'animation:out:await.replace': () => awaitAnimation(),
  'content:replace.before': () => updateBodyClass()
  'content:replace': () => updateBodyClass(),
  'page:view': () => trackPageView()
}

const swup = new Swup({ hooks })
```

**Checks**

### DIFF
--- a/src/Swup.ts
+++ b/src/Swup.ts
@@ -8,7 +8,7 @@ import { type DelegateEventUnsubscribe } from './helpers/delegateEvent.js';
 import { Cache } from './modules/Cache.js';
 import { Classes } from './modules/Classes.js';
 import { type Visit, createVisit } from './modules/Visit.js';
-import { Hooks } from './modules/Hooks.js';
+import { Hooks, type HookName, type HookHandler } from './modules/Hooks.js';
 import { getAnchorElement } from './modules/getAnchorElement.js';
 import { awaitAnimations } from './modules/awaitAnimations.js';
 import { navigate, performNavigation, type NavigationToSelfAction } from './modules/navigate.js';
@@ -43,6 +43,8 @@ export type Options = {
 	linkToSelf: NavigationToSelfAction;
 	/** Enable native animations using the View Transitions API. */
 	native: boolean;
+	/** Hook handlers to register. */
+	hooks: Partial<{ [K in HookName]: HookHandler<K> }>;
 	/** Plugins to register on startup. */
 	plugins: Plugin[];
 	/** Custom headers sent along with fetch requests. */
@@ -61,6 +63,7 @@ const defaults: Options = {
 	animationScope: 'html',
 	cache: true,
 	containers: ['#swup'],
+	hooks: {},
 	ignoreVisit: (url, { el } = {}) => !!el?.closest('[data-no-swup]'),
 	linkSelector: 'a[href]',
 	linkToSelf: 'scroll',
@@ -187,6 +190,12 @@ export default class Swup {
 
 		// Mount plugins
 		this.options.plugins.forEach((plugin) => this.use(plugin));
+
+		// Install user hooks
+		for (const [name, handler] of Object.entries(this.options.hooks)) {
+			// @ts-expect-error: object.entries() seems incapable of preserving key types
+			this.hooks.on(name, handler);
+		}
 
 		// Create initial history record
 		if ((window.history.state as HistoryState)?.source !== 'swup') {
--- a/src/Swup.ts
+++ b/src/Swup.ts
@@ -8,7 +8,7 @@ import { type DelegateEventUnsubscribe } from './helpers/delegateEvent.js';
 import { Cache } from './modules/Cache.js';
 import { Classes } from './modules/Classes.js';
 import { type Visit, createVisit } from './modules/Visit.js';
-import { Hooks, type HookName, type HookHandler } from './modules/Hooks.js';
+import { Hooks, type HookName, type HookInitOptions } from './modules/Hooks.js';
 import { getAnchorElement } from './modules/getAnchorElement.js';
 import { awaitAnimations } from './modules/awaitAnimations.js';
 import { navigate, performNavigation, type NavigationToSelfAction } from './modules/navigate.js';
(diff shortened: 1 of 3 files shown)


hirasso (dismissed):

hirasso:
A few Playwright tests are failing if I run them locally:

CleanShot 2024-07-23 at 13 57 23@2x

daun:
You may have to run `npm run build` before running the tests?

daun:
@hirasso Are the tests passing after building? Let's make sure they're passing on your end before merging :)

hirasso:
I'll check as soon as I find the time

hirasso:
Do you have an idea how supporting `before` could be supported here? I just realised that I need that quite often

daun:
Two ideas.

A — pass in an array: first item is the hook options object, second the handler.

`{ 'content:replace': [{ before: true }, () => {}]}`

B — prefix the hook name

`{ 'before:content:replace': () => {} }`

daun:
They‘re both not amazing, but i like the prefix one better for readability.

daun:
Or `before::content:replace`

Or `before(content:replace)`

Or `before content:replace`

hirasso:
Great examples! I'd also prefer Option B, as with option A it could become hard to have both a before as well as an after-handler, wouldn't it?

hirasso:
> @hirasso Are the tests passing after building? Let's make sure they're passing on your end before merging :)

That was it! After building, the only remaining failing test is animation timing:

CleanShot 2024-07-23 at 19 44 30@2x

hirasso:
> Great examples! I'd also prefer Option B, as with option A it could become hard to have both a before as well as an after-handler, wouldn't it?
> 
> I'd also like to introduce this in this PR already. What do you think?

While I like the obvious `before:content:replace`, the Alpine/Vue-style `content:replace.before` could proof more flexible in the future, if we decide that we want to also add the priority to it: `content:replace.before.priority.300`

daun:
> That was it! After building, the only remaining failing test is animation timing:

Haha, okay, those 2ms strike again.

> I'd also like to introduce this in this PR already. What do you think?

Sure!

> While I like the obvious `before:content:replace`, the Alpine/Vue-style `content:replace.before` could proof more flexible in the future, if we decide that we want to also add the priority to it: `content:replace.before.priority.300`

I think I agree. Both are fine from by perspective, but choosing an established pattern here makes sense. It's also just a simple split by `.` character since we don't use dots in hook names.

daun:
I've added modifiers to the possible options. What's nice is that `replace` comes for free :)

My plan was to allow those in `swup.hooks.on()` as well, but the types immediately became overwhelming and it might not be such a great idea either with the existing options param already doing the same thing.

```js
const hooks = {
  'animation:out:await.replace': () => awaitAnimation(),
  'content:replace.before': () => rememberBodyClass()
  'content:replace': () => updateBodyClass(),
  'page:view': () => trackPageView()
}
```

daun:
What I'm just realizing is that this is a great format for writing simple one-off plugins.

```js
const GsapFadePlugin = {
  'animation:out:await.replace': () => gsap.to('main', { opacity: 0, duration: 0.25 }),
  'animation:in:await.replace': () => gsap.fromTo('main', { opacity: 0 }, { opacity: 1, duration: 0.25 })
}
```

Maybe we could loosen things up there a bit and allow simple objects with hook handlers? Instead of requiring it to extend the base class? We're already removing the handlers on unmount and could do the same here. Exciting 🤠

src/modules/Hooks.ts
```diff
@@ -71,6 +75,12 @@ export type Handlers = {
 	[K in HookName]: HookHandler<K>[];
 };
 
+export type HookInitOptions = {
```
hirasso, line 78: It looks to me like this could be simplified:

```js
export type HookInitOptions = {
	[K in HookName as HookName | HookNameWithModifier]: HookHandler<K>;
};
```

src/Swup.ts
```diff
@@ -188,6 +191,14 @@ export default class Swup {
 		// Mount plugins
 		this.options.plugins.forEach((plugin) => this.use(plugin));
 
+		// Install user hooks
+		for (const [key, handler] of Object.entries(this.options.hooks)) {
+			// Buld options object from hook modifier: 'content:replace.before' => { before: true }
+			const [hook, modifiers] = this.hooks.parseName(key as HookName);
```
hirasso, line 197: This can be simplified, if we use the suggested type simplification down below. Also, then `@ts-expect-error` is not needed anymore

```js
const [hook, modifiers] = this.hooks.parseName(key as HookName | HookNameWithModifier);
this.hooks.on(hook, handler, modifiers);
```

src/Swup.ts
```diff
@@ -188,6 +191,14 @@ export default class Swup {
 		// Mount plugins
 		this.options.plugins.forEach((plugin) => this.use(plugin));
 
+		// Install user hooks
+		for (const [key, handler] of Object.entries(this.options.hooks)) {
+			// Buld options object from hook modifier: 'content:replace.before' => { before: true }
```
hirasso, line 196: "Build..." with an i :)

src/modules/Hooks.ts
```diff
@@ -570,4 +580,14 @@ export class Hooks {
...
 		);
 	}
+
+	/**
+	 * Parse a hook name into the name and any modifiers.
+	 * @param hook Name of the hook.
+	 */
+	parseName(hook: HookName | HookNameWithModifier): [HookName, Partial<HookOptions>] {
+		const [name, ...modifiers] = hook.split('.');
+		const options = modifiers.reduce((acc, mod) => ({ ...acc, [mod]: true }), {});
```
hirasso, line 590: This line is too cryptic for my taste. I'd prefer something like this:

```js
const options = {} as { [key: string]: boolean };
modifiers.forEach((modifier) => {
	options[modifier] = true;
});
```

hirasso (changes requested):
Very nice work!

daun:
@hirasso Great suggestions, will tackle them tomorrow.

src/modules/Hooks.ts
daun: Updated with ``[K in HookName as K | `${K}.${HookModifier}`]: HookHandler<K>`` 🍻

src/Swup.ts
daun: Fixed :)

src/modules/Hooks.ts
daun: Keeping for simple type ergonomics.

daun:
@hirasso Updated as discussed.

hirasso (approved):
💍

daun:
🚀

Issue #84, opened 2018-12-05
Use of the transition-delay property appears to break swup


smedstadc:
Hi, I've been fiddling with swup to evaluate it for a project where it may be useful. So far I think it's pretty cool.

One thing I noticed is the use of the `transition-delay` property to stagger the entrance of elements appears to break page transitions. The page gets stuck in such a state that the user can't navigate and nothing appears.

I ran into this behavior in a bare bones project adapted from the examples in the README. The only additions were a second page to visit, a variant of the `.transition-fade` class with a short delay, and a static navbar that is always visible.

If the user clicks a different navbar link before the transition finishes, swup handles it just fine. If one of the transition classes on the page has a `transition-delay` property assigned to it, then you get the bad behavior I mentioned above.

Is this a limitation of swup, or a bug? 

If you need more information I can put together a clean example.

Thanks!

gmrchk:
Hi @smedstadc, this sounds like a bug in a demo, as swup has no problem with waiting for transition end of delayed elements. Looks like you don't have a "hidden" state defined for at least one of the elements with class `transition-*`. This makes swup wait for the animation to finish, but that animation never even starts. In that case, swup doesn't throw any error (discussed [here](https://github.com/gmrchk/swup/issues/85#issuecomment-445802178)) and the only indication is the debug mode getting stuck at a certain point in the page transition lifecycle.

Of course, maybe I didn't understand the description correctly, and some code or live demo would be awesome if your problem persists.

smedstadc:
Debug mode getting stuck at a point int he transition cycle sounds like what I was seeing. I'll double check my hidden state and see if that was the problem. I'll post an example if it was not.

smedstadc:
After you mentioned the hidden state, I thought that I had forgotten to write a class for the new hidden state or there was a typo in one that I missed because I was looking for more complicated problems. Neither seems to be the case.

I think what happens is when a user starts a page transition, then clicks a different link after the immediate animation starts, but before the delayed animation starts, the cycle can get stuck. This is easier to reproduce with a longer delay.

You can observe the behavior in this project: https://github.com/smedstadc/swup-toy

There still may be something I've missed. I'm just not sure.

gmrchk:
Thank you for the demo, it's really helpful. 

[...]

PR #936, opened 2024-07-23
Manual scroll restoration from history state

daun:
Most routing libraries currently store the scroll position in the history state and manually restore it from there.

**Advantages**

- Programmatically read and manipulate the stored position
- Control over the exact point in time when to restore the scroll position
- No artificial separation between animated and non-animated (history) visits

**Prior art**

- Turbo: [save to history](https://github.com/hotwired/turbo/blob/main/src/core/session.js#L197)
- pjax: [save to history](https://github.com/MoOx/pjax/blob/master/lib/proto/handle-response.js#L23)
- Astro: [save to history](https://github.com/withastro/astro/blob/052d3271d4a9f12b88897c57cdfa134cb83c9eee/packages/astro/src/transitions/events.ts#L168)
- Livewire: [save to body attribute](https://github.com/livewire/livewire/blob/main/js/plugins/navigate/scroll.js#L3)

**Description**

- Remember scroll position in history state
- Restore scroll position from history state
- Completely disable browser scroll restoration
- Create two hooks: `scroll:store` when saving position, `scroll:restore` when restoring position
  - Makes it extendable to support arbitrary scroll containers (to be managed by Scroll Plugin)
- Currently uses debounced `scroll` function to update position when scrolling stops
  - There is the new [scrollend event ](https://developer.mozilla.org/en-US/docs/Web/API/Document/scrollend_event)but it's not 100% supported unfortunately and [the polyfill is huge](https://github.com/argyleink/scrollyfills/blob/main/src/scrollend.js)

**To-do**

- Restore scroll position on reload (maybe a Scroll Plugin feature?)
- Implement proper tests
- Check if this breaks anything and needs to move into v5

**Checks**


daun:
Note to self: make scroll handler passive.

hirasso:
Another note: Check if this breaks scroll position restoration if using `resolveUrl`.

daun:
@hirasso Right! We'll probably need to check this together on a call, the resolve-url logic is still a bit mysterious to me 🙃

daun:
@hirasso I think this one is ready for review :)

Could you check in one of your projects using `resolveUrl` if this makes any difference in behavior?

daun:
Closing in favor of implementing this in the Scroll Plugin for advanced use cases.

daun:
Tracked in [#84](https://github.com/swup/scroll-plugin/issues/84)

PR #960, opened 2024-09-26
Fix hooks replaced once

daun:
**Description**

- The `once` option is currently ignored when replacing an internal hook handler
- This fix allows combinations of `once` and `replace`

**Checks**

### DIFF
--- a/src/modules/Hooks.ts
+++ b/src/modules/Hooks.ts
@@ -77,6 +77,8 @@ export type Handlers = {
 
 export type HookInitOptions = {
 	[K in HookName as K | `${K}.${HookModifier}`]: HookHandler<K>;
+} & {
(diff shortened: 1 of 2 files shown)


hirasso:
Nice! I don't if its a no brainer anyways, but just to be sure: will this also work with 

```js
swup.hooks.replace('enable', () => {}, {once: true});

```
?

daun:
@hirasso Yes, that'll work. The `replace` (and `before` and `once`) methods on the hooks registry passes along any other options. So for example `hooks.replace` looks like this:

```js
replace(hook, handler, options = {}) {
  return this.on(hook, handler, { ...options, replace: true })
}
```

hirasso (approved):

PR #995, opened 2025-04-29
fix: `matchPath`: treat empty array like an empty string

hirasso:
Replaces #994

**Description**

- Transform an empty array to a string before passing it to `match`
- Fixes https://github.com/pillarjs/path-to-regexp/issues/291 without the need for a breaking update to `path-to-regexp`
- Drive-By: Simplify `matchPath` by passing on the argument types using `Parameters<typeof match>`

**Checks**

### DIFF
--- a/src/helpers/matchPath.ts
+++ b/src/helpers/matchPath.ts
@@ -1,19 +1,15 @@
 import { match } from 'path-to-regexp';
 
-import type {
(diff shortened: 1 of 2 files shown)


daun (approved):

daun:
🥨

PR #1005, opened 2025-06-28
Fix link to third-party plugins

hirasso:
**Description**

The link was broken. It's now fixed: https://swup.js.org/third-party-plugins/

Also, rearranged the sentence a bit.

**Checks**

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ Swup is small by design. Extended features can be added via [plugins](https://sw
 - Animate [form submissions](https://swup.js.org/plugins/forms-plugin/)
 - Get help in [debug mode](https://swup.js.org/plugins/debug-plugin/)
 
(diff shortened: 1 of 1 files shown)


daun:
The wording was deliberate to distinguish between plugins (providing new functionality) and integrations (wiring up other services or libraries). Maybe we should update the url in the docs instead? Happy to do it :)

hirasso:
Ah ok, good to know! Sure, let's do that instead :)

hirasso:
Wait... all three listed on the third-party-plugins page are actually plugins. Not so much integrations. Also, "cool URLs don't change" :)

daun:
Oh, right, I was mentally thinking about plugins/integrations in general. We don't do that separation for third-party ones. Good to go then! Sorry for the confusion :)

hirasso:
Great. I'll merge this for now, we can always come back if we feel the need :)

Issue #1022, opened 2026-02-23
`postinstall` script degrades developer experience in PNPM


hirasso:
### What did you expect? 🧐

I'd expect `pnpm add swup` to install swup without warnings

### What actually happened? 😵‍💫

When installing swup via `pnpm`, this warning is printed:

```bash

╭ Warning ───────────────────────────────────────────────────────────────────────────────────╮
│                                                                                            │
│   Ignored build scripts: swup@4.8.2.                                                       │
│   Run "pnpm approve-builds" to pick which dependencies should be allowed to run scripts.   │
│                                                                                            │
╰────────────────────────────────────────────────────────────────────────────────────────────╯
```
This is a new security measure by pnpm – no biggy but it degrades the development experience a bit.

### Swup and plugin versions 📝

- swup: 4.8.2

### Relevant log output 🤓

```shell

```

### URL to minimal reproduction 🔗

n/a

### Checked all these? 📚

- [x] I have read and searched the [official docs](https://swup.js.org)
- [x] I have checked [discussions](https://github.com/orgs/swup/discussions) and [existing issues](https://github.com/issues?q=is:issue+org:swup) for related problems
- [x] I agree to follow this project's [Code of Conduct](https://github.com/swup/.github/blob/main/CODE_OF_CONDUCT.md)
- [x] I have provided all necessary information to the best of my knowledge

daun:
Closed via #1023

PR #1023, opened 2026-02-24
Remove postinstall script from package.json

hirasso:
**Description**

As discussed in #1022 : Remove the `postinstall` script to prevent warnings when installing swup via PNPM.

**Checks**


daun:
There's one flaky test that probably needs a bit more tolerance when measuring animation duration. Or it needs to be rewritten to not measure the duration directly in ms. Usually it works when re-running the job or pushing an empty commit...

daun:
Nice. Didn't know reopening a PR would restart all jobs.

PR #1028, opened 2026-03-01
Create agents file

daun:
**Description**

- Had Copilot create an agents file and tweaked it with some manual corrections
- I've been updating a few things around the tooling and I found it works much better if we have a general high-level document explaining the architecture of Swup and its modules
- Purposefully not CLAUDE.md, but the generic AGENTS.md which all agents should be able to read
- Feel free to push any info you're missing in there :)

**Checks**

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,118 @@
+# Agents
+
+This file provides context for AI coding agents working on the swup codebase.
+
+## Project Overview
+
+**swup** is a versatile and extensible page transition library for server-rendered websites. It manages the complete page load lifecycle, animates between pages, and provides features like caching, smart preloading, native browser history support, and enhanced accessibility.
+
+- **Repository**: [swup/swup](https://github.com/swup/swup)
+- **Documentation**: [swup.js.org](https://swup.js.org)
+- **License**: MIT
+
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -0,0 +1,118 @@
...
+```
+src/
+├── Swup.ts          # Main Swup class
+├── index.ts         # Public exports
+├── helpers.ts       # Re-exports from helpers/
+├── utils.ts         # Re-exports from utils/
+├── config/
+│   └── version.ts   # Library version
+├── helpers/         # Helper functions (classify, delegateEvent, Location, etc.)
+├── modules/         # Core modules (Cache, Hooks, Visit, navigation, animations, etc.)
```
hirasso: Nitpick: casing: Cache, hooks, visit, ... or are we referencing classes here?

hirasso (approved):

AGENTS.md
daun: Cache/Hooks/Visit are classes, and navigation/animations are collections of functions. I do feel like there is a bit of a mixture of OOP vs. Functional going on, but maybe that's not a bad thing?

PR #1029, opened 2026-03-02
Use `actions-up` for pinning actions to commit SHAs

hirasso:
**Description**

Use [`actions-up`](https://github.com/azat-io/actions-up) to update and pin the version of GitHub workflow dependencies in one go.

**Why?**

> With this method, you can point to the specific commit SHA for the desired version of the action. This guarantees that the exact code you've pinned is used, even if the version tag or the action's code is modified later. Commit SHA offers the highest level of security as it tags to a specific point in time within the action's codebase using a unique identifier (the commit SHA).  For example, it could look like "actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332"

– https://www.stepsecurity.io/blog/pinning-github-actions-for-enhanced-security-a-complete-guide

**How**

I opted for a local install via `npm i -D actions-up` and an npm script `{ "actions-up": "actions-up --yes" }` to make the step repeatable with the same `actions-up` version and cli flags.

This PR also already includes the result of running `npm run actions-up`.

**Checks**

- ~New or updated tests are included~
- ~The documentation was updated as required~

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -24,18 +24,18 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@v2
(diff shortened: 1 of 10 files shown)


daun (approved):
Yeah 🍋‍🟩

PR #1036, opened 2026-04-08
Switch to tsdown for bundling

daun:
Move to `tsdown` for bundling.

**Description**

- Move to `tsdown` for bundling the package
- Switch all e2e tests to using the ESM export
- Create an explicit umd test (now called iife according to how tsdown calls it)
- Use identical naming of exports to how plugins do it (`index.js`)

**Benchmarks**

- Build time: **8s** → **1s**
- Bundle size: **7.11 kB** → **7.01 kB** (ESM)
- Bundle size: **8.15 kB** → **7.07 kB** (IIFE/UMD)

**Compatibility**

This should probably go on the `next` branch: the browser compatibility is quite different between both bundlers. Especially the UMD/IIFE build, which was slimmed down by using modern ES syntax there as well.

And a nice excuse to get started on a few planned breaking changes :)

**Checks**


daun:
Tests are passing, ready for review.

I didn't quite notice during testing, but it's obvious when comparing branches: tsdown is superfast! We went from 8s build time to 1s. Bundle sizes are also down.

I'll put this through the astro playground for some more manual testing later this week.

hirasso (approved):
Beautiful! Finally, the CDN version is named correctly IIFE :)

If we want to release this with the next major version, how about we drop the umd version altogether?

daun:
Good point, let's go ESM only 📦

daun:
I just branched a new `next` branch from `main` as base for this PR.

daun:
Merging now so we can get started on v5 🚀

daun:
I've started work on [porting the cli from microbundle to tsdown](https://github.com/swup/cli/pull/61) as well. Needs some more manual testing. But then ideally we're good with a cli update and a major version bump for each plugin, changing output filenames and making them ESM only.

PR #1037, opened 2026-04-09
Ability to ignore visits from hook handlers

### LINKED ISSUE #1008: [Feature Request]: Allow ignoring requests from inside user hooks
### Describe the problem 🧐

In a current project, I need to ignore a navigation depending on the returned content of the new page. The only way to do this currently is to replace an internal handler and throw an exception from there. This forces swup to abandon the navigation and force a browser navigation. Inside normal user hooks, exceptions are only logged and don't lead to a browser navigation.

### Describe the propsed solution 😎

Add a `visit.ignore()` method to every Visit object instance. This would allow ignoring visits at any point. Would either need some validation that ignoring is still possible or a stern warning in the docs to only use the method in the early stages, i.e. before the content is replaced.

This is different from the previously planned and then abandoned `visit.abort()`, which would completely reverse any changes and then stay on the current page. The complexity there was too much to properly implement. Merely ignoring is much easier, it's basically a `window.location = %%`. And we already have all the flags in place to stop the lifecycle in case of aborted visits, which we could re-use here.

### Alternatives considered 🤔

Keep the current solution: replace an internal handler and throw an exception.

### How important is this feature to you? 🧭

Would make my life a lot easier

### Checked all these? 📚

- [x] I have read and searched the [official docs](https://swup.js.org)
- [x] I have checked [discussions](https://github.com/orgs/swup/discussions) and [existing issues](https://github.com/issues?q=is:issue+org:swup) for related problems
- [x] I agree to follow this project's [Code of Conduct](https://github.com/swup/.github/blob/main/CODE_OF_CONDUCT.md)
- [x] I have provided all necessary information to the best of my knowledge

daun:
**Description**

- Allow ignoring visits from hook handlers: `visit.ignore()`
- Ignored visits force a normal browser visit, just like the `ignoreVisit` option
- Useful for ignoring visits based on the actual returned response of the new page
- Closes #1008

**Example**

Ignore a visit (i.e. force a normal browser visit) if the new page has a certain meta tag.

```js
swup.hooks.on('visit:start', (visit) => {
  visit.animation.wait = true
});

swup.hooks.on('page:load', (visit, { page }) => {
  if (page!.html.includes('<meta name="evil-website" content="yes">')) {
    visit.ignore();
  }
});
```

**How it works**

- This relies on the existing try/catch logic for exceptions during navigation
- Exceptions during navigation trigger a back visit, with `skipPopStateHandling` forcing a browser visit
- This change triggers a custom exception for ignored visits
- This is the sanest option here, short of splitting up the file into multiple paths

**Checks**

### DIFF
--- a/src/modules/Visit.ts
+++ b/src/modules/Visit.ts
@@ -172,7 +172,7 @@ export class Visit {
 		return this.state >= VisitState.COMPLETED;
 	}
 
-	/** Was this visit aborted? */
+	/** Was this visit ignored by swup, i.e. left to the browser? */
 	get ignored(): boolean {
 		return this.state === VisitState.IGNORED;
 	}
--- a/src/modules/Visit.ts
+++ b/src/modules/Visit.ts
@@ -86,7 +86,8 @@ export const VisitState = {
 	ENTERING: 6,
 	COMPLETED: 7,
 	ABORTED: 8,
-	FAILED: 9
+	FAILED: 9,
+	IGNORED: 10
 } as const;
 
 /** @internal */
@@ -162,9 +163,18 @@ export class Visit {
 		this.state = VisitState.ABORTED;
 	}
 
-	/** Is this visit done, i.e. completed, failed, or aborted? */
+	ignore() {
+		this.state = VisitState.IGNORED;
+	}
+
+	/** Is this visit done, i.e. completed, aborted, failed, or ignored? */
 	get done(): boolean {
 		return this.state >= VisitState.COMPLETED;
+	}
+
+	/** Was this visit aborted? */
+	get ignored(): boolean {
+		return this.state === VisitState.IGNORED;
 	}
 }
 
--- a/src/modules/navigate.ts
+++ b/src/modules/navigate.ts
@@ -175,6 +175,11 @@ export async function performNavigation(
 			await page;
 		}
 
+		// Ignored in the meantime? Throw to trigger catch block and exit visit
+		if (visit.ignored) {
+			throw new Error(`Visit to ${visit.to.url} manually ignored`);
+		}
+
 		// Check if failed/aborted in the meantime
 		if (visit.done) return;
 
--- a/src/modules/navigate.ts
+++ b/src/modules/navigate.ts
@@ -216,11 +216,11 @@ export async function performNavigation(
 	} catch (error) {
 		// Return early if error is undefined or signals an aborted request
 		if (!error || (error as FetchError)?.aborted) {
-			visit.state = VisitState.ABORTED;
+			visit.advance(VisitState.ABORTED);
 			return;
 		}
 
-		visit.state = VisitState.FAILED;
+		visit.advance(VisitState.FAILED);
 
 		// Log to console
 		console.error(error);
--- a/tests/functional/navigation.spec.ts
+++ b/tests/functional/navigation.spec.ts
@@ -1,6 +1,6 @@
 import { test, expect } from '@playwright/test';
 
-import { clickOnLink, delayRequest, expectToBeAt, sleep } from '../support/commands.js';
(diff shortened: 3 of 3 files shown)


src/modules/navigate.ts
```diff
@@ -175,6 +175,11 @@ export async function performNavigation(
 			await page;
 		}
 
+		// Ignored in the meantime? Throw to trigger catch block and exit visit
+		if (visit.ignored) {
+			throw new Error(`Visit to ${visit.to.url} manually ignored`);
```
daun, line 180: This could also be more cryptic to keep bundle size in check: `throw new Error('ignored')` or so

src/modules/navigate.ts
```diff
@@ -216,11 +221,11 @@ export async function performNavigation(
 	} catch (error) {
 		// Return early if error is undefined or signals an aborted request
 		if (!error || (error as FetchError)?.aborted) {
-			visit.state = VisitState.ABORTED;
+			visit.advance(VisitState.ABORTED);
```
daun, line 224: This change from direct assigment of `visit.state` to the `visit.advance()` method is to make sure we don't change the state of an already finished visit, so ignored stays ignored and doesn't become failed here.

src/modules/navigate.ts
```diff
@@ -216,11 +221,11 @@ export async function performNavigation(
...
 			return;
 		}
 
-		visit.state = VisitState.FAILED;
+		visit.advance(VisitState.FAILED);
```
daun, line 228: This change from direct assigment of visit.state to the visit.advance() method is to make sure we don't change the state of an already finished visit, so ignored stays ignored and doesn't become failed here.

daun:
@hirasso How do you feel about this one? I've needed something like the example above a number of times, and this PR provides an official way without throwing exceptions in replaced hooks 😅

src/modules/navigate.ts
hirasso: I'd prefer clarity here, lets keep it

hirasso:
Sorry, busy times... forgot about this one. I'll look into it!

daun:
There's no specific rush, just figured it fell through the cracks :)

src/modules/Visit.ts
```diff
@@ -162,10 +163,19 @@ export class Visit {
...
+	ignore() {
+		this.state = VisitState.IGNORED;
+	}
+
+	/** Is this visit done, i.e. completed, aborted, failed, or ignored? */
 	get done(): boolean {
 		return this.state >= VisitState.COMPLETED;
 	}
+
+	/** Was this visit aborted? */
```
hirasso: "ignored" instead of "aborted"?

tests/functional/navigation.spec.ts
```diff
@@ -97,4 +97,20 @@ test.describe('navigation', () => {
 		const received = await page.evaluate(() => window.data.received);
 		expect(received).toEqual(expected);
 	});
+
+	test('forces reload for visits ignored from hook', async ({ page }) => {
+		await page.goto('/page-1.html');
+		await waitForSwup(page);
+		await page.evaluate(() => {
+			window._swup.hooks.on('visit:start', (visit) => visit.animation.wait = true);
+			window._swup.hooks.on('page:load', (visit, { page }) => {
```
hirasso, line 106: I don't understand the test, but am sure it's testing what it should :)

hirasso (dismissed):

tests/functional/navigation.spec.ts
daun: It tests that a visit can conditionally be ignored, meaning it's given back for the browser to handle. In this case, it ignores any visit where the new page contains "Page 2" in its html, expecting a normal browser page load to happen.

src/modules/Visit.ts
daun: Oh right 😅

hirasso (approved):

PR #1049, opened 2026-06-04
Unmount all plugins on destroy

daun:
**Description**

- The `destroy` method only unmounts plugins defined in the initial options object
- This will iterate over the actual plugin list instead, including plugins added during runtime

**Checks**


hirasso (approved):
💡

PR #1050, opened 2026-06-08
Switch `CacheData` back to an interface

hirasso:
**Description**

`CacheData` was intended as a way to extend cached data in user land. Only interfaces are extensible/augmentable.

Related discussion: https://github.com/swup/fragment-plugin/pull/101

While working on this, I noticed that `lint:ts` is currently only checking the `src` folder, not the test files. I'll open another PR where this is fixed, including a new type test for the `CacheData` interface.

**Checks**

### DIFF
--- a/src/modules/Cache.ts
+++ b/src/modules/Cache.ts
@@ -2,6 +2,7 @@ import type Swup from '../Swup.js';
 import { Location } from '../helpers.js';
 import { type PageData } from './fetchPage.js';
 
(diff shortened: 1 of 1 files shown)


daun (approved):

hirasso:
Should we release this as a patch together with #1053 as soon as that's merged? I'd say yes :)

daun:
@hirasso Sounds good!

PR #1052, opened 2026-06-08
Lint types in tests

hirasso:
**Description**

- Check test files when running `npm run lint:ts`
- Includes fixes for pre-existing type errors in test files
- Includes a new test for `CacheData` being augmentable via declaration merging

**Checks**

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "dev": "npm run clean && BROWSERSLIST_ENV=development microbundle src/index.ts -w -f modern",
     "format": "npx eslint . --fix",
     "lint": "npm run lint:ts && npm run lint:es",
-    "lint:ts": "tsc --noEmit --skipLibCheck",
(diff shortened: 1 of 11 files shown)


daun (commented):
Nice. Is there a specific reason we need a separate tsconfig for the tests? Ideally we could keep the same config.

hirasso:
I can try to change the default tsconfig tomorrow to support both `src` as well as `tests`

hirasso:
I don't know why I was hesitating to simply include everything in the main `tsconfig.json`. Anyways, now the main tsconfig is responsible for both, `src` and `test` :)

daun:
Nice! One less file :)

The hesitancy might come from having to change `rootDir` from `src` to `.`. Feels a bit dangerous hehe 🤠

daun (approved):

PR #1053, opened 2026-06-09
Avoid collisions in hook ids

daun:
**Description**

- The next hook handler's id is currently derived from the hook ledger's count
- Removing a hook leads to duplicated hook ids and handlers that are not removed
- This PR changes it to always increment, avoiding collisions

**Checks**

### DIFF
--- a/src/modules/Hooks.ts
+++ b/src/modules/Hooks.ts
@@ -168,6 +168,8 @@ export class Hooks {
 		'visit:end'
 	];
 
+	protected nextHookId = 0;
+
 	constructor(swup: Swup) {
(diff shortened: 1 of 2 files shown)


hirasso (approved):
Nice! How did you find out about this?

daun:
@hirasso I was debugging a hook that wasn't being removed with `.off()`. It turned out to be something else (a stale closure/function reference) but it felt plausible that this might cause issues in other scenarios so we'd better fix it anyway 🤠 The original idea of using the ledger size as next id came from the research into event/hook systems, but it probably was a library without an `off()` method.

PR #1061, opened 2026-06-16
Go ESM only

daun:
**Description**

- ESM only
- Remove CJS and IIFE builds
- Keep a special ESM bundle with all dependencies inlined for e2e tests
- Upgrade tsdown to 0.22 → upgrades rolldown to v1 stable
- The dts file renaming hack after build is no longer required
- Clean fixtures dist folder before tests to make sure outdated dist files are ignored in tests

**package.json keys**

- Current build tools check `exports.import` before `module`, so it's safe to remove & easy to add back if necessary
- Build tools also check `exports.types`, but the root-level `types` key is required for the TS badge on npm :)

**Checks**


hirasso (approved):
🚀

